Issue #6, opened 2023-07-21
Feature: ability to `bake` with many repos or a `--file` flag with a number of repos


jpmcb:
### Type of feature

🍕 Feature

### Current behavior

Currently, multiple calls are required for in order to `bake` multiple repos.

### Suggested solution

We should allow users to provide a list of repos when calling `bake` and define a `--file` flag that has a list of many repos in it to send to the pizza-oven service.

So, for multiple repos on the command line, that would look like:

```
$ pizza bake https://github.com/open-sauced/pizza https://github.com/open-sauced/pizza-cli ... etc.
```

With a file, the file definition might look like:
```yaml
repos:
    - https://github.com/open-sauced/pizza
    - https://github.com/open-sauced/pizza-cli
    - ... etc.
```

And that could be provided by:

```
$ pizza bake --file my-repos.yaml
```

The tricky thing will be getting the UX of this all right but should lead to a slick experience for doing big batch jobs with many repos.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

k1nho:
I would like to investigate this one! is there an expected behavior for say this case

```bash
$ pizza bake https://github.com/open-sauced/pizza https://github.com/open-sauced/pizza-cli ... --file myrepos.yaml
```

jpmcb:
@k1nho - good question!!!

In the case that the user provides _both_ the file flag and repo URLs, the CLI should parse both. We may run into some odd edge cases where there are overlaps but from a user experience perspective, I can see the need to have a "base" file and add some other repos directly on the CLI.

jpmcb:
Closed by #12

PR #12, opened 2023-07-26
Implements the option to batch requests to /bake

k1nho:
## Description

This PR implements the option to batch requests to /bake which can be achieved 2 ways:
1) Variadic number of URLs, i.e, ```pizza bake https://github.com/open-sauced/pizza https://github.com/open-sauced/pizza-cli```
2) yaml file, i.e, ```pizza bake --file=config.yaml```, or using shorthand ```pizza bake -f=config.yaml```

If both options are provided, the command merges all requests and avoids duplicate requests by storing URLs in a map prior to sending the request.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

#6 ability to bake many repos 

## Added tests?

- [ ] 👍 yes
- [x] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/cmd/bake/bake.go
+++ b/cmd/bake/bake.go
@@ -9,8 +9,11 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 
 	"github.com/spf13/cobra"
+
+	"gopkg.in/yaml.v3"
 )
 
 // Options are the options for the pizza bake command including user
@@ -19,16 +22,19 @@ type Options struct {
 	// Endpoint is the service endpoint to reach out to
 	Endpoint string
 
-	// URL is the git repo URL that will be sourced via 'pizza bake'
-	URL string
+	// URLs are the git repo URLs that will be sourced via 'pizza bake'
+	URLs []string
 
 	// Wait defines the client choice to wait for /bake to finish processing
 	Wait bool
+
+	// FilePath is the location of the file containing a batch of repos to be baked
+	FilePath string
 }
 
 const bakeLongDesc string = `WARNING: Proof of concept feature.
 
-The bake command takes a URL to a git repository and uses a pizza-oven service
+The bake command accepts one or multiple URLs to a git repository and uses a pizza-oven service
 to source those commits. These commits will then be used for insights on OpenSauced.`
 
 // NewBakeCommand returns a new cobra command for 'pizza bake'
@@ -40,16 +46,13 @@ func NewBakeCommand() *cobra.Command {
 		Short: "Use a pizza-oven to source git commits into OpenSauced",
 		Long:  bakeLongDesc,
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) > 1 {
-				return errors.New("only a single url can be ingested at a time")
-			}
-			if len(args) == 0 {
-				return errors.New("must specify the URL of a git repository")
+			if len(args) == 0 && opts.FilePath == "" {
+				return errors.New("must specify the URL(s) of a git repository or provide a batch file")
 			}
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			opts.URL = args[0]
+			opts.URLs = append(opts.URLs, args...)
 			return run(opts)
 		},
 	}
@@ -58,6 +61,7 @@ func NewBakeCommand() *cobra.Command {
 	// For now, localhost is fine.
 	cmd.Flags().StringVarP(&opts.Endpoint, "endpoint", "e", "http://localhost:8080", "The endpoint to send requests to")
 	cmd.Flags().BoolVarP(&opts.Wait, "wait", "w", false, "Wait for bake processing to finish")
+	cmd.Flags().StringVarP(&opts.FilePath, "file", "f", "", "The yaml file containing a series of repos to batch to /bake")
 
 	return cmd
 }
@@ -67,19 +71,62 @@ type bakePostRequest struct {
 	Wait bool   `json:"wait,omitempty"`
 }
 
+type repos struct {
+	URLs []string `yaml:"repos"`
+}
+
 func run(opts *Options) error {
-	bodyPostReq := &bakePostRequest{
-		URL:  opts.URL,
-		Wait: opts.Wait,
+	var repos repos
+	uniqueURLs := make(map[string]bool)
+
+	if opts.FilePath != "" {
+		configFile, err := os.ReadFile(opts.FilePath)
+		if err != nil {
+			return err
+		}
+
+		err = yaml.Unmarshal(configFile, &repos)
+		if err != nil {
+			return err
+		}
+
+		for _, url := range repos.URLs {
+			uniqueURLs[url] = true
+		}
 	}
 
+	// make sure there are no duplicated queries to the same URL
+	for _, url := range opts.URLs {
+		if _, ok := uniqueURLs[url]; !ok {
+			uniqueURLs[url] = true
+			continue
+		}
+		fmt.Printf("Warning: duplicated URL (%s) would not be processed again\n", url)
+	}
+
+	for url := range uniqueURLs {
+		bodyPostReq := bakePostRequest{
+			URL:  url,
+			Wait: opts.Wait,
+		}
+
+		err := bakeRepo(bodyPostReq, opts.Endpoint)
+		if err != nil {
+			fmt.Printf("Error: failed fetch of %s repository (%s)\n", url, err.Error())
+		}
+	}
+
+	return nil
+}
+
+func bakeRepo(bodyPostReq bakePostRequest, endpoint string) error {
 	bodyPostJSON, err := json.Marshal(bodyPostReq)
 	if err != nil {
 		return err
 	}
 
 	responseBody := bytes.NewBuffer(bodyPostJSON)
-	resp, err := http.Post(fmt.Sprintf("%s/bake", opts.Endpoint), "application/json", responseBody)
+	resp, err := http.Post(fmt.Sprintf("%s/bake", endpoint), "application/json", responseBody)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


bdougie:
Assuming someone already has the list of repo urls in a CSV or spreadsheet ready to paste, this option works?

```
pizza bake \
https://github.com/open-sauced/pizza \
https://github.com/open-sauced/pizza-cli
```

But would it would be nice UX if they did need to add `\`

```
pizza bake |
https://github.com/open-sauced/pizza 
https://github.com/open-sauced/pizza-cli
https://github.com/open-sauced/docs
https://github.com/open-sauced/insights
```

or 

```
pizza bake --list
https://github.com/open-sauced/pizza 
https://github.com/open-sauced/pizza-cli
https://github.com/open-sauced/docs
https://github.com/open-sauced/insights
```
Not what is possible, but open to suggestions or push back.

k1nho:
```
pizza bake \
https://github.com/open-sauced/pizza \
https://github.com/open-sauced/pizza-cli
```
Yep, this option falls under variadic number of URLs and is supported at the moment.

jpmcb:
> Assuming someone already has the list of repo urls in a CSV or spreadsheet ready to paste, this option works?

If they are space delimited, this would work out of the box:

```sh
$ pizza bake https://github.com/open-sauced/pizza https://github.com/open-sauced/pizza-cli ...
```

But the user would _have to_ escape the newline with the `\` if they are newline delimited. There's not a good way to get around this without breaking traditional linux UX and convention. I would not recommend we implement something that expects newlines on the command line. For example, providing a `--list` flag that expects a new line delimited list of things, this wouldn't really work within a script or interpreted by `bash`, `zsh`, etc.

For this use case, typically, 'd recommend someone use xargs, sed, awk, or other linux tooling that solves this problem through good I/O and piping practices:

```sh
$ cat my-repos.csv | xargs -I {} pizza bake {}
```

If this UX is too rough, then providing the `--file` flag as proposed in #6 will enable people to smooth that out.

go.mod
```diff
@@ -11,4 +11,5 @@ require (
 require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	golang.org/x/sys v0.9.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
```
jpmcb, line 14: Make sure to run `go mod tidy`.

This isn't an indirect dependency since you're using it explicitly in `bake.go` and the `go` tooling should bump it up to the top level require block automatically.

cmd/bake/bake.go
```diff
@@ -9,8 +9,10 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
```
jpmcb, line 15: I typically like to seperate blocks of imports by their starting import path (i.e., group all the `github.com` ones together, group all the `gopkg` ones together, etc.). Something like this should suffice:

```suggestion
	"github.com/spf13/cobra"

	"gopkg.in/yaml.v3"
```

cmd/bake/bake.go
```diff
@@ -19,11 +21,14 @@ type Options struct {
 	// Endpoint is the service endpoint to reach out to
 	Endpoint string
 
-	// URL is the git repo URL that will be sourced via 'pizza bake'
-	URL string
+	// URLs are the git repo URLs that will be source via 'pizza bake'
+	URLs []string
```
jpmcb, line 25: The `bakeLongDesc` usage string also needs to be updated to reference taking multiple urls (not just one)

cmd/bake/bake.go
```diff
@@ -19,11 +21,14 @@ type Options struct {
 	// Endpoint is the service endpoint to reach out to
 	Endpoint string
 
-	// URL is the git repo URL that will be sourced via 'pizza bake'
-	URL string
+	// URLs are the git repo URLs that will be source via 'pizza bake'
```
jpmcb, line 24: nit: The grammar here is off

```suggestion
	// URLs are the git repo URLs that will be sourced via 'pizza bake'
```

cmd/bake/bake.go
```diff
@@ -40,16 +45,13 @@ func NewBakeCommand() *cobra.Command {
 		Short: "Use a pizza-oven to source git commits into OpenSauced",
 		Long:  bakeLongDesc,
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) > 1 {
-				return errors.New("only a single url can be ingested at a time")
-			}
-			if len(args) == 0 {
+			if len(args) == 0 && opts.FilePath == "" {
 				return errors.New("must specify the URL of a git repository")
```
jpmcb, line 49: This error string needs updating:

```suggestion
				return errors.New("must specify the URL(s) of a git repository or provide a batch file")
```

cmd/bake/bake.go
```diff
@@ -67,19 +70,74 @@ type bakePostRequest struct {
...
 func run(opts *Options) error {
-	bodyPostReq := &bakePostRequest{
-		URL:  opts.URL,
-		Wait: opts.Wait,
+	var repos repos
+	uniqueURLs := make(map[string]int)
+
+	if opts.FilePath != "" {
+		configFile, err := os.ReadFile(opts.FilePath)
+
```
jpmcb, line 88: nit: Unnecessary newline spacing here. Typically error blocks appear right after where the error was generated in Go.

cmd/bake/bake.go
```diff
@@ -67,19 +70,74 @@ type bakePostRequest struct {
...
+		for _, url := range repos.URLs {
+			uniqueURLs[url] = 0
+		}
 	}
 
+	// make sure there are no duplicated queries to the same URL
+	for _, url := range opts.URLs {
+		if _, exists := uniqueURLs[url]; !exists {
+			uniqueURLs[url] = 0
+		}
```
jpmcb, line 108: The typical pattern for checking access on a hashmap is:

```suggestion
		if _, ok := uniqueURLs[url]; !ok {
			uniqueURLs[url] = 0
		}
```

cmd/bake/bake.go
```diff
@@ -67,19 +70,74 @@ type bakePostRequest struct {
...
+
+		for _, url := range repos.URLs {
+			uniqueURLs[url] = 0
+		}
 	}
 
+	// make sure there are no duplicated queries to the same URL
+	for _, url := range opts.URLs {
+		if _, exists := uniqueURLs[url]; !exists {
+			uniqueURLs[url] = 0
```
jpmcb, line 107: If we are going to make this optimization checking for duplicates, we should log a warning to the user that a duplicate url was found and it won't be processed again.

cmd/bake/bake.go
```diff
@@ -67,19 +70,74 @@ type bakePostRequest struct {
...
+	URL string
+	Err error
+}
+
 func run(opts *Options) error {
-	bodyPostReq := &bakePostRequest{
-		URL:  opts.URL,
-		Wait: opts.Wait,
+	var repos repos
+	uniqueURLs := make(map[string]int)
```
jpmcb, line 84: I'm not sure about using a `map[string]int` over `map[string]bool`. Really, it's just a fast access list and we don't really care about the values, just the keys.

Maybe there's a more straight forward datastructure for this in Go?

cmd/bake/bake.go
```diff
@@ -67,19 +70,74 @@ type bakePostRequest struct {
...
+
+	for url := range uniqueURLs {
+		bodyPostReq := bakePostRequest{
+			URL:  url,
+			Wait: opts.Wait,
+		}
+
+		err := bakeRepo(bodyPostReq, opts.Endpoint)
+		batchStatus = append(batchStatus, bakeStatus{URL: url, Err: err})
+	}
```
jpmcb, line 121: A future optimization would be to put this all behind a `go func` so that we can batch all the requests at once. The server is able to handle many requests and this would be a good optimization instead of processing each one by one.

But this is fine for now, we can make an issue for that after the fact.

cmd/bake/bake.go
```diff
@@ -67,19 +70,74 @@ type bakePostRequest struct {
...
+
+	for _, status := range batchStatus {
+		if status.Err != nil {
+			fmt.Printf("Failed: %v\n", status.Err)
+		} else {
+			fmt.Printf("%s: Success!\n", status.URL)
+		}
+	}
+	return nil
+}
```
jpmcb, line 131: I would opt to just log the failure immediately after it's received by the server: this block will be confusing without the context of _which_ repo is being processed.

cmd/bake/bake.go
k1nho: Changed it to ```map[string]bool```, makes more sense.

> Maybe there's a more straight forward datastructure for this in Go?
Not that I know of, wished Go had sets 😂, but the map will do.

cmd/bake/bake.go
```diff
@@ -95,4 +142,5 @@ func run(opts *Options) error {
 	}
 
 	return nil
+
```
jpmcb, line 145: nit: Unnecessary new line

jpmcb (approved):
Nicely done here!!! 👏🏼

cmd/bake/bake.go
jpmcb: Yeah, `map[type]bool` is about as close as we can get 🥲

bdougie (approved):
I appreciate the work on this and even the providing some invalidation to my approaches in the issue :)

PR #15, opened 2023-08-01
feat: repo-query support

### LINKED ISSUE #14: feat: Repo query command line interface chat
The [repo-query](https://github.com/open-sauced/repo-query/issues/3) is an engine that powers asking questions about a git repository.

As a fast, proof of concept, let's get something in the pizza CLI that uses the repo-query service.

Some scaffolding user experience:

```
$ pizza repo-query https://github.com/open-sauced/insights --branch beta

Indexing ...

Complete!

Want to ask a question?
> How is the highlights button getting its color?

The highlights button is getting its color from the CSS classes applied to it in
the highlights-filter-card.tsx file. Specifically, the button has the class bg-light-slate-1,
which sets the background color, and text-light-slate-12, which sets the text color.
These classes are defined in the CSS file associated with the component.

Want to ask another question?
>
```

_Originally posted by @jpmcb in https://github.com/open-sauced/repo-query/issues/3#issuecomment-1650609768_

diivi:
## Description
Fixes #14 

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

#14 

Peek 2023-08-01 23-28

Implemented a loop for QnA after I recorded this GIF.

## Added tests?

- [ ] 👍 yes
- [x] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [x] 🙅 no documentation needed

### DIFF
--- a/cmd/repo-query/repo-query.go
+++ b/cmd/repo-query/repo-query.go
@@ -206,7 +206,7 @@ func processIndexChunk(chunk string) {
 	chunkLines := strings.Split(chunk, "\n")
 	eventLine := chunkLines[0]
 	event := strings.Split(eventLine, ": ")[1]
-	
+
 	switch event {
 	case "FETCH_REPO":
 		fmt.Println("Fetching Repository from GitHub...")
--- a/cmd/repo-query/repo-query.go
+++ b/cmd/repo-query/repo-query.go
@@ -131,9 +131,8 @@ func run(opts *Options) error {
 		}
 
 		fmt.Printf("An error occurred: %v\n", string(body))
+		return errors.New("error while checking if repo is indexed")
 	}
-
-	return nil
 }
 
 type indexPostRequest struct {
@@ -168,13 +167,17 @@ func indexRepo(owner string, repo string, branch string) error {
 		}
 
 		fmt.Printf("An error occurred: %v\n", string(body))
+		return errors.New("error while indexing repository")
 	}
 
 	// listen for SSEs and send data,event pairs to processIndexChunk
 	reader := bufio.NewReader(resp.Body)
 	for {
 		line, err := reader.ReadString('\n')
-		if err != nil {
+		// if err is eof then break
+		if err == io.EOF {
+			return nil
+		} else if err != nil {
 			return err
 		}
 
@@ -197,12 +200,15 @@ func indexRepo(owner string, repo string, branch string) error {
 				chunk += line
 			}
 
-			processIndexChunk(chunk)
+			err := processIndexChunk(chunk)
+			if err != nil {
+				return err
+			}
 		}
 	}
 }
 
-func processIndexChunk(chunk string) {
+func processIndexChunk(chunk string) error {
 	chunkLines := strings.Split(chunk, "\n")
 	eventLine := chunkLines[0]
 	event := strings.Split(eventLine, ": ")[1]
@@ -215,12 +221,15 @@ func processIndexChunk(chunk string) {
 	case "SAVE_EMBEDDINGS":
 		fmt.Println("Saving the embeddings to our database...")
 	case "ERROR":
-		fmt.Println("There was an error while indexing this repository. Redirecting to the Home Page.")
+		fmt.Println("There was an error while indexing this repository.")
+		return errors.New("error while indexing repository")
 	case "DONE":
 		fmt.Println("Indexing Complete. You can now ask questions about this repository! 🎉")
 	default:
 		break
 	}
+
+	return nil
 }
 
 type queryPostRequest struct {
@@ -265,13 +274,17 @@ func askQuestion(question string, owner string, repo string, branch string) erro
 		}
 
 		fmt.Printf("An error occurred: %v\n", string(body))
+		return errors.New("error while asking question")
 	}
 
 	//  listen for SSEs and send data,event pairs to processChatChunk
 	reader := bufio.NewReader(resp.Body)
 	for {
 		line, err := reader.ReadString('\n')
-		if err != nil {
+		// if err is eof then break
+		if err == io.EOF {
+			return nil
+		} else if err != nil {
 			return err
 		}
 
@@ -294,12 +307,15 @@ func askQuestion(question string, owner string, repo string, branch string) erro
 				chunk += line
 			}
 
-			processChatChunk(chunk)
+			err := processChatChunk(chunk)
+			if err != nil {
+				return err
+			}
 		}
 	}
 }
 
-func processChatChunk(chunk string) {
+func processChatChunk(chunk string) error {
 	chunkLines := strings.Split(chunk, "\n")
 	eventLine := chunkLines[0]
 	dataLine := chunkLines[1]
@@ -328,7 +344,10 @@ func processChatChunk(chunk string) {
 		fmt.Println(data)
 	case "ERROR":
 		fmt.Println("Something went wrong. Please try again.")
+		return errors.New("error while asking question")
 	default:
 		break
 	}
+
+	return nil
 }
--- a/cmd/repo-query/repo-query.go
+++ b/cmd/repo-query/repo-query.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -118,12 +119,26 @@ func run(opts *Options) error {
 		fmt.Println("Repo found ✅")
 
 		for {
+			// if ctrl+c is pressed, exit
+			c:= make(chan os.Signal, 1)
+			signal.Notify(c, os.Interrupt)
+
+			go func() {
+				<-c
+				fmt.Println("\n🍕Exiting...");
+				os.Exit(0)
+			}()
+
 			fmt.Printf("\nWant to ask a question about %s/%s?\n", owner, repo)
 			fmt.Printf("> ")
 			// read input
 			scanner := bufio.NewScanner(os.Stdin)
 			if scanner.Scan() {
 				input := scanner.Text()
+				if input == "exit" {
+					fmt.Println("🍕Exiting...")
+					os.Exit(0)
+				}
 				err := askQuestion(input, owner, repo, opts.branch)
 				if err != nil {
 					return err
--- a/cmd/repo-query/repo-query.go
+++ b/cmd/repo-query/repo-query.go
@@ -369,6 +369,9 @@ func processChatChunk(chunk string) error {
 		fmt.Println("Generating a response...🧠")
 	case "DONE":
 		fmt.Println()
+		// the server sends the data as a string, with \n, "", and \t escaped.
+		replacer := strings.NewReplacer("\\n", "\n", "\\\"", "\"", "\\t", "\t")
+		data := replacer.Replace(data.(string))
 		fmt.Println(data)
 	case "ERROR":
 		fmt.Println("Something went wrong. Please try again.")
--- a/cmd/repo-query/repo-query.go
+++ b/cmd/repo-query/repo-query.go
@@ -55,7 +55,11 @@ func NewRepoQueryCommand() *cobra.Command {
 	return cmd
 }
 
-func getOwnerAndRepo(url string) (owner, repo string) {
+func getOwnerAndRepo(url string) (owner, repo string, err error) {
+	if !strings.HasPrefix(url, "https://github.com/") {
+		return "", "", fmt.Errorf("invalid URL: %s", url)
+	}
+
 	// Remove the "https://github.com/" prefix from the URL
 	url = strings.TrimPrefix(url, "https://github.com/")
 
@@ -67,17 +71,18 @@ func getOwnerAndRepo(url string) (owner, repo string) {
 		owner = segments[0]
 		repo = segments[1]
 	} else {
-		// URL is not in the expected format
-		owner = ""
-		repo = ""
+		return "", "", fmt.Errorf("invalid URL: %s", url)
 	}
 
-	return owner, repo
+	return owner, repo, nil
 }
 
 func run(opts *Options) error {
 	// get repo name and owner name from URL
-	owner, repo := getOwnerAndRepo(opts.URL)
+	owner, repo, err := getOwnerAndRepo(opts.URL)
+	if err != nil {
+		return err
+	}
 
 	fmt.Printf("Checking if %s/%s is indexed by us...⏳\n", owner, repo)
 	resp, err := http.Get(fmt.Sprintf("%s/collection?owner=%s&name=%s&branch=%s", repoQueryURL, owner, repo, opts.branch))
@@ -85,8 +90,9 @@ func run(opts *Options) error {
 		return err
 	}
 
-	// not found or ok or error
-	if resp.StatusCode == http.StatusNotFound {
+	switch resp.StatusCode {
+	case http.StatusNotFound:
+		// repo is not indexed
 		fmt.Println("Repo not found❗")
 		fmt.Println("Indexing repo...⏳")
 		err := indexRepo(owner, repo, opts.branch)
@@ -107,10 +113,10 @@ func run(opts *Options) error {
 				}
 			}
 		}
-	} else if resp.StatusCode == http.StatusOK {
+	case http.StatusOK:
+		// repo is indexed
 		fmt.Println("Repo found ✅")
 
-		// this should be an infinite loop
 		for {
 			fmt.Printf("\nWant to ask a question about %s/%s?\n", owner, repo)
 			fmt.Printf("> ")
@@ -124,7 +130,7 @@ func run(opts *Options) error {
 				}
 			}
 		}
-	} else {
+	default:
 		body, err := io.ReadAll(resp.Body)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


diivi:
This is mostly done (except reviews), but there's this error that I get after a repo gets indexed:

```
Checking if pawurb/ecto_psql_extras is indexed by us...⏳
Repo not found❗
Indexing repo...⏳
Fetching Repository from GitHub...
Embedding Repository...
Saving the embeddings to our database...
Indexing Complete. You can now ask questions about this repository! 🎉
Error: EOF
Usage:
  pizza repo-query url [flags]

Flags:
  -b, --branch string   The branch to index (default "HEAD")
  -h, --help            help for repo-query

2023/08/02 00:25:50 EOF
exit status 1
```
here's the code for sending the indexing request:
https://github.com/diivi/pizza-cli/blob/57d49f5b1cb41a29ba10ecfbc81b77e3399d30f2/cmd/repo-query/repo-query.go#L145-L146
Can you help me understand what might be wrong here?

diivi:
Never mind, I fixed it.

bdougie:
I noticed that but after I asked a question, I got exit 1 error. 

```
go run main.go repo-query https://github.com/open-sauced/insights --branch beta
Checking if open-sauced/insights is indexed by us...⏳
Repo found ✅

Want to ask a question about open-sauced/insights?
> what is the test plan?
Searching the codebase for your query...🔍
Searching tests/lib/utils/truncate.test.ts for your query...🔍
Generating a response...🧠

The test plan for the repository \"open-sauced/insights\" can be found in the file \"tests/lib/utils/truncate.test.ts\". Here are some relevant test cases from that file:\n\n1. Test case: \"string should be returned unchanged\"\n   - Test code: `test(\"string should be returned unchanged\", () => { const string = \"test/testtesttest\"; const result = truncateString(string, 17); expect(result).toBe(\"test/testtesttest\"); });`\n\n2. Test case: \"string should truncate with length\"\n   - Test code: `test(\"string should truncate with length\", () => { const string = \"test/testtesttest\"; const result = truncateString(string, 9); expect(result).toBe(\"test/test...\"); });`\n\nPlease note that these are just a few examples. There might be more test cases in the test plan
Error: EOF
Usage:
  pizza repo-query url [flags]

Flags:
  -b, --branch string   The branch to index (default "HEAD")
  -h, --help            help for repo-query

2023/08/01 11:59:53 EOF
exit status 1
```

diivi:
> I noticed that but after I asked a question, I got exit 1 error.
Fixed

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// The first segment is the owner, and the second segment is the repository name
+	if len(segments) >= 2 {
+		owner = segments[0]
+		repo = segments[1]
+	} else {
+		// URL is not in the expected format
+		owner = ""
+		repo = ""
+	}
```
jpmcb, line 73: Is this case at all valid? This would make sense to return an error here that can be handled by the caller.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+	owner, repo := getOwnerAndRepo(opts.URL)
+
+	fmt.Printf("Checking if %s/%s is indexed by us...⏳\n", owner, repo)
+	resp, err := http.Get(fmt.Sprintf("%s/collection?owner=%s&name=%s&branch=%s", repoQueryURL, owner, repo, opts.branch))
+	if err != nil {
+		return err
+	}
+
+	// not found or ok or error
+	if resp.StatusCode == http.StatusNotFound {
```
jpmcb, line 89: This is a good use case for a tagged switch off of `resp.StatusCode`. Something like:

```go
    switch resp.StatusCode {
    case http.StatusNotFound:
        // handle the repo not being found
    case http.StatusOK:
        // handle the repo being found
    }
```

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+		Branch: branch,
+	}
+
+	indexPostJSON, err := json.Marshal(indexPostReq)
+	if err != nil {
+		return err
+	}
+
+	responseBody := bytes.NewBuffer(indexPostJSON)
+	resp, err := http.Post(fmt.Sprintf("%s/embed", repoQueryURL), "application/json", responseBody)
```
jpmcb, line 157: This chunk can be simplified since that bytes buffer won't be needed latter:

```suggestion
        resp, err := http.Post(fmt.Sprintf("%s/embed", repoQueryURL), "application/json", bytes.NewBuffer(indexPostJSON))
```

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("An error occurred: %v\n", string(body))
+		return errors.New("error while indexing repository")
```
jpmcb, line 170: The error handling throughout could be abit more structured. I'd suggest using `fmt.Errorf` and returning a detailed error:

```suggestion
		if err != nil {
			return fmt.Errorf("error reading response body: %s", err.Error())
		}

		return fmt.Errorf("error while indexing repository: %s", err.Error())
```

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	// listen for SSEs and send data,event pairs to processIndexChunk
+	reader := bufio.NewReader(resp.Body)
+	for {
```
jpmcb, line 175: This whole chunk from 175 through 209 seems really brittle.

But maybe I'm not understanding it fully, let me see if I have it right: essentially, we're anticipating getting server side events where:

- EOF from the response body indicates the server events are done
- new lines can be thrown away and are essentially empty events from the server.
- Anything that has `event: ` indicates a new event from the server and we should process the string chunks up to (and including the newline).
  - A empty newline string in this chunk indicates that event is done (?)

Maybe this all just needs some more inline documentation since it's not immediately clear _how_ this is supposed to work or how it's supposed to interact with the server. Even better, we could write a clear "handler" that has individual methods for each scenario. But that may be overkill: regardless, we should harden this since I anticipate this going stale really quickly.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+				return err
+			}
+		}
+	}
+}
+
+func processIndexChunk(chunk string) error {
+	chunkLines := strings.Split(chunk, "\n")
+	eventLine := chunkLines[0]
+	event := strings.Split(eventLine, ": ")[1]
```
jpmcb, line 214: This is also rather brittle but probably just needs some inline documentation on what's going on.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+		},
+	}
+
+	queryPostJSON, err := json.Marshal(queryPostReq)
+	if err != nil {
+		return err
+	}
+
+	responseBody := bytes.NewBuffer(queryPostJSON)
+	resp, err := http.Post(fmt.Sprintf("%s/query", repoQueryURL), "application/json", responseBody)
```
jpmcb, line 264: Same here: could simplify by using the buffer directly:

```suggestion
	resp, err := http.Post(fmt.Sprintf("%s/query", repoQueryURL), "application/json", bytes.NewBuffer(queryPostJSON))
```

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+			return err
+		}
+
+		fmt.Printf("An error occurred: %v\n", string(body))
+		return errors.New("error while asking question")
+	}
+
+	//  listen for SSEs and send data,event pairs to processChatChunk
+	reader := bufio.NewReader(resp.Body)
+	for {
```
jpmcb, line 282: Looking at this chunk on lines 282 through 316, it looks nearly identical to the above chunk for processing server side events: let's abstract this away and reduce code duplication.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+				return err
+			}
+		}
+	}
+}
+
+func processChatChunk(chunk string) error {
+	chunkLines := strings.Split(chunk, "\n")
+	eventLine := chunkLines[0]
+	dataLine := chunkLines[1]
```
jpmcb, line 321: And same here: worth some docs and maybe abstracting away to reduce code duplication.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+
+	event := strings.Split(eventLine, ": ")[1]
+	var data interface{} // the data can be a string or a JSON object
```
jpmcb, line 324: > the data can be a string or a JSON object

Hmmm that's unfortunate. Can this returned data be normalized somehow? Might be worth ensuring the server side is returning normalized data.

What you have is fine, but attempting to cast `data.(map[string]interface{})` will panic if this is not a `map[string]interface{}` type. I.e., if this is indeed a string for some reason, this will break and panic mid flight.

It looks like the only case where `data` is a string is during `DONE`. Can that be normalized on the server side? Might need @Anush008's comments too.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+			// read input
+			scanner := bufio.NewScanner(os.Stdin)
+			if scanner.Scan() {
+				input := scanner.Text()
+				err := askQuestion(input, owner, repo, opts.branch)
+				if err != nil {
+					return err
+				}
+			}
+		}
```
jpmcb, line 109: It'd be really nice if this could handle exit scenarios: i.e., if the user enters `exit` or hits `<CTRL+C>`, it'd be nice if the chat client gracefully shut down.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	cmd.Flags().StringVarP(&opts.branch, "branch", "b", "HEAD", "The branch to index")
+
+	return cmd
+}
+
+func getOwnerAndRepo(url string) (owner, repo string) {
+	// Remove the "https://github.com/" prefix from the URL
+	url = strings.TrimPrefix(url, "https://github.com/")
```
jpmcb, line 60: This is abit brittle. Entering anything other the `https://github.com/...` produces a confusing error:

```
❯ go run main.go repo-query github.com/open-sauced/pizza
Checking if github.com/open-sauced is indexed by us...⏳
Repo not found❗
Indexing repo...⏳
An error occurred: Unable to fetch repository license
Error: error while indexing repository
Usage:
  pizza repo-query url [flags]

Flags:
  -b, --branch string   The branch to index (default "HEAD")
  -h, --help            help for repo-query

2023/08/01 15:37:03 error while indexing repository
exit status 1
```

We should maybe look for a good regex that captures most requests to GitHub. We should probably also add that only github.com is supported (for now) in the help message.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,353 @@
...
+
+	return owner, repo
+}
+
+func run(opts *Options) error {
+	// get repo name and owner name from URL
+	owner, repo := getOwnerAndRepo(opts.URL)
+
+	fmt.Printf("Checking if %s/%s is indexed by us...⏳\n", owner, repo)
+	resp, err := http.Get(fmt.Sprintf("%s/collection?owner=%s&name=%s&branch=%s", repoQueryURL, owner, repo, opts.branch))
```
jpmcb, line 83: I wonder if someone could inject query parameters into this URL that would inevitably end up being pulled down by the repo-query service. We should think deeply about the security posture of this client and should do every reasonable effort to harden against those sorts of injections or attacks.

cmd/repo-query/repo-query.go
diivi: Yup, I had to handle some of these cases (empty events, inconsistent responses etc.)   here too - https://github.com/open-sauced/ai/pull/226/files#diff-34c2ff6688d75e301c74751cff9760478c59b33e97c949f5b05dfe1f273517d2R94. I just had the context already so I pasted things in here. Will document this.

cmd/repo-query/repo-query.go
diivi: Yeah, he said he'll normalise it, but what you said is what we have currently (string on `DONE` and json for other events)

cmd/repo-query/repo-query.go
diivi: Hm, aren't these things supposed to be handled by the backend ideally?

jpmcb: In theory, they should be. But client side validation doesn't hurt. We can always open another issue after this proof of concept merges too.

diivi:
I added some parsing errors and some documentation here and there. Also abstracted the SSE listening code.
Let me know if there's anything more than I can improve.

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,381 @@
...
+
+The repo-query command takes a URL to a git repository and indexes it so
+users can ask questions about it.`
+
+func NewRepoQueryCommand() *cobra.Command {
+	opts := &Options{}
+
+	cmd := &cobra.Command{
+		Use:   "repo-query url [flags]",
+		Short: "Ask questions about a git repository",
```
diivi, line 37:
```suggestion
		Short: "Ask questions about a GitHub repository",
```

bdougie:
I think this is in a place to test live, but only after we switch the Azure deployment. https://github.com/open-sauced/repo-query/issues/10

It would also be nice to format the response return a little syntax highlighting. Not sure if that is a repo-query thing (via system prompt asking for syntax highlight) or something we can do with the response in cobra. Here is the output in markdown:

```
> can you convert hsl(30, 70.0%, 7.2%) to hex code?
Searching the codebase for your query...🔍
Searching components/molecules/PaginationGotoPage/pagination-goto-page.tsx for your query...🔍
Generating a response...🧠

Yes, there is a function called `hslToHex` in the codebase. It is used in the `components/molecules/PaginationGotoPage/pagination-goto-page.tsx` file. Here is an example of how it is used:\n\n```tsx\nconst PaginationGotoPage = ({ page, name, setPage, totalPage }: PaginationGotoPageProps): JSX.Element => {\n  const [pageNumber, setPageNumber] = useState<number | string>(page);\n  const [value, setValue] = useState<number | string>(\"\");\n\n  const handleGotoPage = (e: React.FormEvent<HTMLFormElement>) => {\n    e.preventDefault();\n    // Convert HSL to Hex using the hslToHex function\n    const hexColor = hslToHex(30, 70.0, 7.2);\n    // Rest of the code...\n  }\n\n  // Rest of the code...\n}\n```\n\nPlease note that the exact implementation of the `hslToHex` function is not provided in the code snippet. You may need to search for its implementation in other files or external dependencies
``` 

Here is a screencap from the terminal:

Screen Shot 2023-08-02 at 9 16 13 AM

jpmcb:
> It would also be nice to format the response return a little syntax highlighting

@bdougie - I was thinking the exact same thing. Yes, this is something we can absolutely tackle and would be trivial with the `fmt` package:

```go
package main

import (
	"fmt"
)

func main() {
	myString := "This is a string\n\nwith newlines"
	fmt.Println(myString)
}
```

this prints a the string as:
```
This is a string

with newlines
```

As far as deeper syntax highlighting (like with backticks ` ``` ` or other markdown style editing from the AI model, that would be more difficult since we'd need to inspect the user's `TERM` env variable and ensure the terminal environment supports 256 colors.

But for now, interpreting the newlines as literal would be a great first step.

jpmcb:
> For the "\n"s, it's a known issue, the server sends "\\n", so this gets rendered as a normal string. The markdown parser for the open-sauced/ai interface has the same problem (the prompt needs to fix this server-side).

@diivi - I'd use a string replacer then:

```go
package main

import (
	"fmt"
	"strings"
)

func main() {
	myString := "This is a string\\n\\nwith newlines"
	replacer := strings.NewReplacer(
		"\\n", "\n",
	)
	myString = replacer.Replace(myString)
	fmt.Println(myString)
}
```

this returns the correct bits for the string to be formated by `fmt`

cmd/repo-query/repo-query.go
```diff
@@ -0,0 +1,384 @@
...
+	for {
+		// if ctrl+c is pressed, exit
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt)
+
+		go func() {
+			<-c
+			fmt.Println("\n🍕Exiting...")
+			os.Exit(0)
+		}()
```
jpmcb, line 139: Very cool solution!! 👏🏼 🍕

jpmcb (approved):
I think overall this looks good 👍🏼 I'm sure there will be things to iterate on: for example, I'm hoping to integrate a fancier logging library that was made for CLI applications #17

bdougie (approved):
I like the approach and look forward to seeing how this evolves.

PR #21, opened 2023-08-15
feat: cli auth

### LINKED ISSUE #20: Feature: `pizza login` to get user's bearer token
### Type of feature

🍕 Feature

### Current behavior

Many of the OpenSauced API endpoints require bearer authentication. We should add a `pizza login` command that gets the user's bearer token, stores it (much like a cookie), and can be used for various operations.

This will enable us to do some pretty interesting things with the Pizza CLI (like get a user's highlights, getting their dashboards, etc.)

### Suggested solution

1. A command can be called to log a user in with `pizza login`
2. This command should open a new browser tab. This should be our authentication endpoint (supabase?) and should have a `?redirect=` back to a locally started Go web server in order to capture the bearer token
3. Once the user is logged in and the redirect is completed, the bearer token should be stored within a `~/.pizza/` directory. This is essentially like storing a bearer token cookie in the browser

### Additional context

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

Anush008:
## Description
This PR adds authentication to the application.
As mentioned in #20, the auth info is written to `~/.pizza/session.json`.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Resolves #20.

pizza-cli-auth

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook
- [x] 🙅 no documentation needed

### DIFF
--- a/cmd/auth/auth.go
+++ b/cmd/auth/auth.go
@@ -1,0 +1,191 @@
+package auth
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"crypto/sha256"
+	_ "embed"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"os/signal"
+	"path"
+	"time"
+
+	"github.com/pkg/browser"
+	"github.com/spf13/cobra"
+)
+
+const (
+	codeChallengeLength = 87
+	supabaseID          = "ibcwmlhcimymasokhgvn"
+	supabasePublicKey   = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoiYW5vbiIsImlhdCI6MTYyOTkzMDc3OCwiZXhwIjoxOTQ1NTA2Nzc4fQ.zcdbd7kDhk7iNSMo8SjsTaXi0wlLNNQcSZkzZ84NUDg"
+	authCallbackAddr    = "localhost:3000"
+	sessionFileName     = "session.json"
+)
+
+//go:embed success.html
+var successHtml string
+
+func NewLoginCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "login",
+		Short: "Log into the CLI application via GitHub",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return run()
+		},
+	}
+
+	return cmd
+}
+
+func run() error {
+	codeVerifier, codeChallenge, err := pkce(codeChallengeLength)
+	if err != nil {
+		return fmt.Errorf("PKCE error: %v", err)
+	}
+
+	supabaseAuthURL := fmt.Sprintf("https://%s.supabase.co/auth/v1/authorize", supabaseID)
+	queryParams := url.Values{
+		"provider":              {"github"},
+		"code_challenge":        {codeChallenge},
+		"code_challenge_method": {"S256"},
+		"redirect_to":           {"http://" + authCallbackAddr + "/"},
+	}
+
+	authenticationURL := supabaseAuthURL + "?" + queryParams.Encode()
+
+	server := &http.Server{Addr: authCallbackAddr}
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		defer shutdown(server)
+
+		code := r.URL.Query().Get("code")
+		if code == "" {
+			http.Error(w, "Error: auth-code query param not found", http.StatusBadRequest)
+			return
+		}
+
+		sessionData, err := getSession(code, codeVerifier)
+		if err != nil {
+			http.Error(w, "Error: Access token exchange failed", http.StatusInternalServerError)
+			return
+		}
+
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			fmt.Println("Error getting home directory:", err)
+			return
+		}
+
+		dirName := path.Join(homeDir, ".pizza")
+		if err := os.MkdirAll(dirName, os.ModePerm); err != nil {
+			http.Error(w, ".pizza directory couldn't be created", http.StatusInternalServerError)
+			return
+		}
+
+		jsonData, err := json.Marshal(sessionData)
+		if err != nil {
+			http.Error(w, "Error marshaling session data", http.StatusInternalServerError)
+			return
+		}
+
+		filePath := path.Join(dirName, sessionFileName)
+		if err := os.WriteFile(filePath, jsonData, 0o600); err != nil {
+			http.Error(w, "Error writing to file", http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "text/html; charset=utf-8")
+		w.Write([]byte(successHtml))
+
+		username := sessionData["user"].(map[string]interface{})["user_metadata"].(map[string]interface{})["user_name"].(string)
+		fmt.Println("🎉 Login successful 🎉")
+		fmt.Println("Welcome aboard", username, "🍕")
+
+	})
+
+	err = browser.OpenURL(authenticationURL)
+	if err != nil {
+		fmt.Println("Failed to open the browser 🤦‍♂️")
+		fmt.Println("Navigate to the following URL to begin authentication:")
+		fmt.Println(authenticationURL)
+	}
+
+	errCh := make(chan error)
+	go func() {
+		errCh <- server.ListenAndServe()
+	}()
+
+	interruptCh := make(chan os.Signal, 1)
+	signal.Notify(interruptCh, os.Interrupt)
+
+	select {
+	case err := <-errCh:
+		if err != nil && err != http.ErrServerClosed {
+			return err
+		}
+	case <-time.After(60 * time.Second):
+		shutdown(server)
+		return errors.New("authentication timeout")
+	case <-interruptCh:
+		fmt.Println("\nAuthentication interrupted❗️")
+		shutdown(server)
+		os.Exit(0)
+	}
+	return nil
+}
+
+func getSession(authCode, codeVerifier string) (map[string]interface{}, error) {
+	url := fmt.Sprintf("https://%s.supabase.co/auth/v1/token?grant_type=pkce", supabaseID)
+
+	payload := map[string]string{
+		"auth_code":     authCode,
+		"code_verifier": codeVerifier,
+	}
+
+	jsonPayload, _ := json.Marshal(payload)
+
+	req, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonPayload))
+	req.Header.Set("Content-Type", "application/json;charset=UTF-8")
+	req.Header.Set("ApiKey", supabasePublicKey)
+
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status: %s", res.Status)
+	}
+
+	var responseData map[string]interface{}
+	if err := json.NewDecoder(res.Body).Decode(&responseData); err != nil {
+		return nil, err
+	}
+
+	return responseData, nil
+}
+
+func pkce(length int) (verifier, challenge string, err error) {
+	p := make([]byte, length)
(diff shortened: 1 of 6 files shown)


jpmcb:
Wow this is looking great: fast turnaround. Good work. Overall looks good. Doing a deep dive review now

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/pkg/browser"
+	"github.com/spf13/cobra"
+)
+
+const (
+	codeChallengeLength = 87
+	supabaseID          = "ibcwmlhcimymasokhgvn"
+	supabasePublicKey   = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoiYW5vbiIsImlhdCI6MTYyOTkzMDc3OCwiZXhwIjoxOTQ1NTA2Nzc4fQ.zcdbd7kDhk7iNSMo8SjsTaXi0wlLNNQcSZkzZ84NUDg"
+	authCallbackAddr    = "localhost:3000"
+	sessionFileName     = "session.json"
```
jpmcb, line 30: This is probably going to need to be exported so other commands can determine the session filename: maybe drop this into a `pkg/constants.go` file that other commands can utilize?

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func pkce(length int) (verifier, challenge string, err error) {
+	p := make([]byte, length)
+	if _, err := io.ReadFull(rand.Reader, p); err != nil {
+		return "", "", err
+	}
+	verifier = base64.RawURLEncoding.EncodeToString(p)
+	b := sha256.Sum256([]byte(verifier))
+	challenge = base64.RawURLEncoding.EncodeToString(b[:])
```
jpmcb, line 188: Isn't `b[:]` unnecessary here? Isn't that the same as just `b` being passed in to the encoding function?

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+	return verifier, challenge, nil
+}
+
+func shutdown(server *http.Server) {
+	go func() {
+		err := server.Shutdown(context.Background())
+		if err != nil {
+			fmt.Println("Graceful shutdown failed", err)
```
jpmcb: This error message could be abit better. And may make sense to use `panic` here since the graceful shutdown should almost never happen.

go.mod
```diff
@@ -11,5 +11,6 @@ require (
 
 require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
```
jpmcb, line 14: Hmmm I'm abit hesitent using anything from `github.com/pkg` although it's fine for this proof of concept. Many of those projects haven't gotten updates in years and this is pulling in a pretty old version of `golang.org/x/sys` which carries some CVEs.

Looks like this was forked into `github.com/cli/browser` which is being maintained by github. Can we use that one instead?

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+
+		sessionData, err := getSession(code, codeVerifier)
+		if err != nil {
+			http.Error(w, "Error: Access token exchange failed", http.StatusInternalServerError)
+			return
+		}
+
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			fmt.Println("Error getting home directory:", err)
```
jpmcb: This should be `fmt.Errorf` with the error attached.

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+		return "", "", err
+	}
+	verifier = base64.RawURLEncoding.EncodeToString(p)
+	b := sha256.Sum256([]byte(verifier))
+	challenge = base64.RawURLEncoding.EncodeToString(b[:])
+	return verifier, challenge, nil
+}
+
+func shutdown(server *http.Server) {
+	go func() {
```
jpmcb, line 193: This shouldn't need to be in its own go func. When this is called, I'd imagine the caller would want to wait for the shutdown to happen.

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+		return errors.New("authentication timeout")
+	case <-interruptCh:
+		fmt.Println("\nAuthentication interrupted❗️")
+		shutdown(server)
+		os.Exit(0)
+	}
+	return nil
+}
+
+func getSession(authCode, codeVerifier string) (map[string]interface{}, error) {
```
jpmcb, line 149: `map[string]interface{}` is really general. Is the shape of this json well known? We should create `type` structs for these to be decoded into.

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+
+	jsonPayload, _ := json.Marshal(payload)
+
+	req, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonPayload))
+	req.Header.Set("Content-Type", "application/json;charset=UTF-8")
+	req.Header.Set("ApiKey", supabasePublicKey)
+
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
```
jpmcb, line 165: These should be wrapped in `fmt.Errorf` so error tracability is easier. Something like:

```go
fmt.Errorf("could not make req with default client: %s", err.Error())
```

README.md
```diff
@@ -5,13 +5,17 @@ This CLI can be used for all things OpenSauced!
 ```
 ❯ pizza
 
+A command line utility for insights, metrics, and all things OpenSauced
```
jpmcb, line 8: 🍕

go.mod
bdougie: I agree with using what the cli/cli uses. We can count on them to also keep an eye out for updates and upstream alongside us.

Anush008: https://github.com/open-sauced/pizza-cli/pull/21/commits/5a0b17de099522911138dd11169abf3c22888b22

cmd/auth/auth.go
Anush008: I got the response schema from Supabase's OpenAPI spec. https://github.com/supabase/gotrue/blob/4ff1fe058cfab418c445808004091e89dcf87124/openapi.yaml#L1881
Created structs representative of the schema.
https://github.com/open-sauced/pizza-cli/pull/21/commits/7d0ce543cf4762dad0ec5f37262adbb718501d20

cmd/auth/auth.go
Anush008: I tried doing that at first.
But without the go func,  the server doesn't seem to wait for the existing request to complete before closing it's sockets, which cuts the auth-flow abruptly. Even when the call deferred. 
https://github.com/Anush008/pizza-cli/blob/7d0ce543cf4762dad0ec5f37262adbb718501d20/cmd/auth/auth.go#L65-L66.

cmd/auth/auth.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			fmt.Println("Error getting home directory:", err)
+			return
+		}
+
+		dirName := path.Join(homeDir, ".pizza")
+		if err := os.MkdirAll(dirName, os.ModePerm); err != nil {
+			http.Error(w, ".pizza directory couldn't be created", http.StatusInternalServerError)
+			return
+		}
```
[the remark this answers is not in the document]
Anush008: I had a handler at first. But later found that `MkdirAll` does nothing if the path is already a dir. Which is what we'd expect.
https://cs.opensource.google/go/go/+/go1.21.0:src/os/path.go;l=18

cmd/auth/auth.go
Anush008: https://github.com/open-sauced/pizza-cli/commit/5a0b17de099522911138dd11169abf3c22888b22

cmd/auth/auth.go
Anush008: `sha256.Sum256([]byte(verifier))` returns a `[32]byte`, but `EncodeToString(b[:])` requires a byte slice. So the `b[:]`.

jpmcb: Huh interesting! Good catch!!

cmd/auth/auth.go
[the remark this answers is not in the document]
jpmcb: Great! 👏🏼

pkg/constants.go
```diff
@@ -0,0 +1,9 @@
+package constants
```
jpmcb, line 1: This breaks typical go convention where the package being imported is the same as the name of the directory it's in.

And almost all of these are specific to the auth package, except for `session.json` which is a common file name other commands may need for gaining the session token.

My suggestion:
- move most of these into a `cmd/auth/constants.go` file under the auth package
- move this file to `pkg/constants/constants.go` with the "session.json" variable exposed

There's probably a dozen ways to do this, so if you think there's a cleaner way, feel free to suggest something!

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 )
 
 require (
+	github.com/cli/browser v1.2.0 // indirect
```
jpmcb, line 13: This is a direct dependency. You need to run `go mod tidy` to clean up dependencies in the go.mod / go.sum

cmd/auth/auth.go
```diff
@@ -0,0 +1,239 @@
...
+	UpdatedAt              string                 `json:"updated_at"`
+	DeletedAt              string                 `json:"deleted_at"`
+}
+
+type MFAFactorSchema struct {
+	ID           string `json:"id"`
+	Status       string `json:"status"`
+	FriendlyName string `json:"friendly_name"`
+	FactorType   string `json:"factor_type"`
+}
```
jpmcb, line 239: Very nice: thanks for getting these all interpolated out.

Usually structs appear towards the top of a file, or at least before the functions that use them. Feel free to drop these in a different file within the auth package if that makes sense too.

These all also probably make sense as private structs to the auth package: I doubt we'll need these exported by `auth` package consumers:

```go
type accessTokenResponse struct {
```

cmd/auth/auth.go
```diff
@@ -0,0 +1,239 @@
...
+	constants "github.com/open-sauced/pizza-cli/pkg"
+	"github.com/spf13/cobra"
+)
+
+//go:embed success.html
+var successHTML string
+
+const loginLongDesc string = `Log into the application.
+
+This command initiates the GitHub auth flow to log you into the application by launching your browser`
```
jpmcb, line 31: nit: Wording

```suggestion
const loginLongDesc string = `Log into OpenSauced.

This command initiates the GitHub auth flow to log you into the OpenSauced application by launching your browser`
```

cmd/auth/auth.go
```diff
@@ -0,0 +1,239 @@
...
+		return fmt.Errorf("PKCE error: %v", err.Error())
+	}
+
+	supabaseAuthURL := fmt.Sprintf("https://%s.supabase.co/auth/v1/authorize", constants.SupabaseID)
+	queryParams := url.Values{
+		"provider":              {"github"},
+		"code_challenge":        {codeChallenge},
+		"code_challenge_method": {"S256"},
+		"redirect_to":           {"http://" + constants.AuthCallbackAddr + "/"},
+	}
```
jpmcb, line 59: Question: I tried abit of an experiment where I logged out of my page on `insights.opensauced.pizza` and did `pizza login`. It automatically logged me in without opening the github auth flow. I suppose the same thing happens when hitting the "Connect with GitHub" button in the insights UI; it just automatically logs me back in.

Anush008: Yes. You won't see the GitHub auth page everytime and will log in automatically.

cmd/auth/auth.go
Anush008: https://github.com/open-sauced/pizza-cli/pull/21/commits/6f02e2c54f6f3700195dd842760b067b5fdc29de

bdougie (approved):
Looks great. Love the colors and logo on the splash page

Issue #9, opened 2023-07-21
Feature: `pizza show <git-repo-url>` for a realtime dashboard of repo activity


jpmcb:
### Type of feature

🍕 Feature

### Current behavior

Via the CLI, there's not a good way to show a realtime view of events happening on a repository.

### Suggested solution

We should implement a `show` command that takes a git repo URL and displays the realtime events (commits) for that repo:

```
$ pizza show <git-repo-url>
```

This could use some Terminal-User-Interface (TUI) to makeup a nice looking display.

The use-case here is maybe an office with a centralized monitor where people want to display a terminal with these realtime metrics of a repo they are watching.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

bdougie:
Perhaps there are some existing libraries or examples we can borrow from for the visualizations. I shared the [nba-go](https://github.com/homerchen19/nba-go) as my favorite example, but also citing for others. 

Also the charmcli project might have some interesting visualizations in their org or in the the works. Happy to reach out if a collaboration is of interest.

k1nho:
> Also the charmcli project might have some interesting visualizations in their org or in the the works. Happy to reach out if a collaboration is of interest.

Yep, lots of the Charm libs use Go. [BubbleTea](https://github.com/charmbracelet/bubbletea), [Bubbles](https://github.com/charmbracelet/bubbles), and [Lipgloss](https://github.com/charmbracelet/lipgloss) are the ones I've seen used the most in combination to create TUI's

k1nho:
@jpmcb Do we know the fields to be displayed for this?  and are we are requesting to pizza-oven for some commit data periodically after some seconds to be realtime ? I think we may be able to leverage Charm Bubbles table component structure for this. I can explore this one.

bdougie:
If we have commits data already we can visualize commits over a time period in a line graph, similar to how we do in the UI.

k1nho:
>New contributors for a repo: https://api.opensauced.pizza/#/Contributors%20service/newPullRequestContributors
Churn contributors for a repo: https://api.opensauced.pizza/#/Contributors%20service/findAllChurnPullRequestContributors

It seems these two endpoints are broken atm, just returning 500 status. ```v1/contributors/insights/recent``` and ```v1/contributors/search``` seem to be the only two working.

jpmcb:
[...]

PR #24, opened 2023-08-16
feat: Pizza show

k1nho:
## Description
This PR implements the pizza show command. This is the first step in developing a TUI that can be used for a centralized monitor to display metrics for an specific repository.

Current metrics:
* Repository model
   - name, issues, forks, stars, size
   - New contributors table
   - Churned contributors table
* Contributor model (interactive through repository model tables)
   - pr velocity, pr count, issues opened, is maintainer
   - The last 5 PRs made
   
wip

https://github.com/open-sauced/pizza-cli/assets/59541661/1689182b-cab8-4663-81f0-1a8c9c7b9bc9

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

#9 

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


jpmcb:
👏🏼 👏🏼 👏🏼 👏🏼 👏🏼 👏🏼 👏🏼

cmd/show/show.go
```diff
@@ -0,0 +1,211 @@
...
+
+type showRespose struct {
+	Contributors []Author `json:"data"`
+	Metadata     Metadata `json:"meta"`
+}
+
+type reposResponse struct {
+	ID                int       `json:"id"`
+	Size              int       `json:"size"`
+	Issues            int       `json:"issues"`
```
jpmcb, line 85: Incredible work so far, you've been crushing it!!

I've been wondering if we should create some sort of generated Go package that makes working with the OpenSauced API easier.

This is good for now, but I worry about drift happening between these inline structs and API changes we make. cc @brandonroberts - I'll look at if there's anything possible with generating Go structs and APIs from swagger or NextJS definitions.

cmd/show/show.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return nil
+}
+
+// FetchNewContributors: Returns all the new contributors
+func FetchNewContributors(opts *Options, repoName string, repoID int) (showRespose, error) {
+	var client = &http.Client{Timeout: 10 * time.Second}
+	var newContributors showRespose
+
+	resp, err := client.Get(fmt.Sprintf("https://api.opensauced.pizza/v1/contributors/insights/new?page=%d&limit=%d&range=%d&topic=javascript&repo=%s&repoIds=%d",
```
brandonroberts, line 197: The `topic` and `repo` params aren't needed here. Fixed in https://github.com/open-sauced/api/pull/265

```suggestion
	resp, err := client.Get(fmt.Sprintf("https://api.opensauced.pizza/v1/contributors/insights/new?page=%d&limit=%d&range=%d&repoIds=%d",
```

jpmcb (changes requested):
FYI: this will be blocked on getting this done https://github.com/open-sauced/pizza-cli/issues/28

which should drastically reduce the boilerplate in the CLI here we'd have to maintain 👍🏼

k1nho:
That would be awesome! if it wasn't for GoTags I would still be typing the structs 😂. Btw, any preference on the color palette? I picked #FF4500 which seems to be what is used in the UI, but Lipgloss has [adaptive colors](https://github.com/charmbracelet/lipgloss#adaptive-colors) which can be leveraged depending on the terminal background color. Also, since I grab the repo info I can also render the size, issues and so on, anything from there that may be useful to display? Lastly, is there an endpoint to get the commits from the repo_id ? I see there is one from pizza to get it from the bake repo id.

jpmcb:
> Btw, any preference on the color palette? I picked #FF4500 which seems to be what is used in the UI, but Lipgloss has [adaptive colors](https://github.com/charmbracelet/lipgloss#adaptive-colors) which can be leveraged depending on the terminal background color.

Yes, we should 100% use adaptive colors: a hard coded color value will be really rough for people on random weird color schemes.

> Also, since I grab the repo info I can also render the size, issues and so on, anything from there that may be useful to display?

Good question: it's pretty open ended right now, so if something makes sense to add, go for it!

> Lastly, is there an endpoint to get the commits from the repo_id ? I see there is one from pizza to get it from the bake repo id.

Yeah, good question, I was just chatting with @brandonroberts about this: there isn't currently a way to get the commits from the API based on a `repo_id`. Right now, `baked_repos` and `repos` are sort of siloed and we don't have a great way in the bakend to connect them.

The first thing to do here is get our ETL service to connect repo ids in the `baked_repos` table that are null. And if it can't find them, then go and fetch that data from github.

So, for now, commits are abit blocked, but it's ok to leave that out for now.

cmd/show/show.go
jpmcb: @k1nho - The generated Go API has landed 👀 

https://github.com/open-sauced/go-api

Let's try using that and open issues in that repo with any sharp edges you might hit!

jpmcb:
Can you change this PR's target branch to be `beta`? 👀 We recently changed the default branch to `beta` in support of using the releaser action

For further reading:
- https://github.com/open-sauced/pizza-cli/issues/34

jpmcb:
@k1nho - let me know when you'd like me to look at this 😄 happy to help if you need anything along the way

k1nho:
I think is looking good for the 1st iteration of ```pizza show```, updated the demo at the top to reflect what I have now. Still need to do some clean up some of the error handling between cobra<-TUI and TUI<-models when doing async operations. Also, for now I just have the static data from first fetch on dashboard model, so 'reatime' is not there yet, but the contributor model is dynamic so that one is 'realtime'. @jpmcb would appreciate a review of what I have so far.

cmd/show/dashboard.go
```diff
@@ -0,0 +1,323 @@
...
+
+	go func() {
+		wg.Wait()
+		close(doneChan)
+	}()
+
+	select {
+	case <-doneChan:
+		break
+	case err := <-errorChan:
```
jpmcb, line 89: I think I gave similar feedback to @cecobask (and I'm probably misremembering how these channels work) but won't this only take the first error and squash any other errors that land in this channel? We should surface any error that might happen.

cmd/show/dashboard.go
```diff
@@ -0,0 +1,323 @@
...
+		switch msg.String() {
+		case "right", "l":
+			m.tableView = (m.tableView + 1) % 2
+		case "left", "h":
+			if m.tableView-1 <= 0 {
+				m.tableView = 0
+			} else {
+				m.tableView--
+			}
+		case "q", "ctrl+c":
```
jpmcb, line 114: People might also hit `<ESC>` to quit. And probably `ctrl-d` (which you see alot in interpreters etc)

cmd/show/contributors.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+	return m, cmd
+}
+
+func (m ContributorModel) View() string {
+	if m.userInfo != nil {
+		return m.drawContributorView()
+	}
+	return "🚧  Not found"
```
jpmcb, line 70: This was all I got when hitting enter on some alumni contributors and then it seemed to freeze and not take anymore input. Maybe this is a non-starter return or blocks somehow?

cmd/show/dashboard.go
```diff
@@ -0,0 +1,323 @@
...
+	title6 := lipgloss.JoinHorizontal(lipgloss.Left, string(titleRunes6))
+	title := lipgloss.JoinVertical(lipgloss.Center, title1, title2, title3, title4, title5, title6)
+	titleView := lipgloss.NewStyle().Width(m.newContributorsTable.Width()).Margin(2, 60).Foreground(color()).Render(title)
+
+	return titleView
+}
+
+// drawRepositoryTitle: view of the repository info (name, stars, size, and issues)
+func (m DashboardModel) drawRepositoryInfo() string {
+	repoName := lipgloss.NewStyle().Bold(true).AlignHorizontal(lipgloss.Center).Render(fmt.Sprintf("Repository: %s", m.RepositoryInfo.FullName))
```
jpmcb, line 170: This line paniced when i tried `pizza show freecodecamp/freecodecamp`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0xc0 pc=0x10071735c]

goroutine 25 [running]:
github.com/open-sauced/pizza-cli/cmd/show.DashboardModel.drawRepositoryInfo({{{{{...}, {...}, 0x0}, {{...}, {...}, 0x0}, {{...}, {...}, 0x0}, {{...}, ...}, ...}, ...}, ...})
        /Users/jpmcb/workspace/opensauced/pizza-cli/cmd/show/dashboard.go:170 +0x13c
github.com/open-sauced/pizza-cli/cmd/show.DashboardModel.drawDashboardView.func2()
        /Users/jpmcb/workspace/opensauced/pizza-cli/cmd/show/dashboard.go:220 +0x7c
created by github.com/open-sauced/pizza-cli/cmd/show.DashboardModel.drawDashboardView in goroutine 1
        /Users/jpmcb/workspace/opensauced/pizza-cli/cmd/show/dashboard.go:218 +0x180
```

It also didn't want to return control back to my terminal ... which is weird since the program panicked out.That might be unique to my setup tho since I'm nested deep in tmux usually too so maybe `ctrl-c` was being eaten up somewhere else.

jpmcb (commented):
Any way to make this abit more dynamic? I don't know "tea" very well but it'd be awesome if this was abit more responsive to the height/width of the given terminal.

Screenshot 2023-08-30 at 7 55 07 PM

---

Overall looks like you're on the right track and you already have alot!!! Like you said, just some rough edges to iron out.

If you're looking for a pattern to implement some more "real" live views of the data, `time.Timer` is a really good pattern I've seen for this kind of thing: https://blogtitle.github.io/go-advanced-concurrency-patterns-part-2-timers/

cmd/show/tui.go
```diff
@@ -0,0 +1,87 @@
...
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+// ██████╗ ██╗███████╗███████╗ █████╗
+// ██╔══██╗██║╚══███╔╝╚══███╔╝██╔══██╗
+// ██████╔╝██║  ███╔╝   ███╔╝ ███████║
+// ██╔═══╝ ██║ ███╔╝   ███╔╝  ██╔══██║
+// ██║     ██║███████╗███████╗██║  ██║
+// ╚═╝     ╚═╝╚══════╝╚══════╝╚═╝  ╚═╝
```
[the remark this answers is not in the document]
k1nho: 😆 oops had it there as reference when making the title view

cmd/show/tui.go
```diff
@@ -0,0 +1,87 @@
...
+func pizzaTUI(opts *Options) {
+
+	// TODO: handle error to cobra
+	dashboardModel, _ := InitDashboard(opts)
+
+	model := MainModel{dashboard: dashboardModel}
+
+	if _, err := tea.NewProgram(model, tea.WithAltScreen()).Run(); err != nil {
+		fmt.Println("Error running program:", err)
+		os.Exit(1)
```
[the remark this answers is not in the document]
k1nho: Ok, yeah this makes sense. Normally for bubbletea only programs that is the standard way, but here since we are interfacing with cobra it makes sense to bubble the error to it.

cmd/show/dashboard.go
k1nho: Yep, that is correct only the first error will be shown the rest is squashed, was not sure if we were going for the errgroup pattern for these type of concurrent operations, but makes sense to surface all errors.

cmd/show/contributors.go
k1nho: Interesting, usually when this hits 'Not found' you can go back to the dashboard model with 'backspace' maybe adding the help menu could be useful for navigation ? I've seen this pattern in other bubbletea programs. I am restructuring this to avoid this nil checking on the contributor and only render valid contributors by and giving a quick response via the dashboard model like this.

Screenshot 2023-09-01 at 1 30 18 PM

cmd/show/dashboard.go
k1nho: Yep, this will indeed panic as it tries to load a repo that is not indexed by the api, and then dereferences nil when loading the dashboard model. This is part of the error flow, handling this in the upcoming changes.

> It also didn't want to return control back to my terminal ... which is weird since the program panicked out.That might be unique to my setup tho since I'm nested deep in tmux usually too so maybe ctrl-c was being eaten up somewhere else.

Yeah, this could be a problem with these setups, which is why bubbletea programs often have 'q', 'esc', and other keymaps to terminate. I know 'esc' gets eaten out when I'm using Vim terminal, so something similar could be happening with tmux.

cmd/show/dashboard.go
```diff
@@ -0,0 +1,360 @@
...
+
+// validateShowQuery: validates fields set to query the contributor tables
+func validateShowQuery(opts *Options) error {
+	if opts.RepoName == "" {
+		return errors.New("no repository was provided")
+	}
+
+	if len(strings.Split(opts.RepoName, "/")) != 2 {
+		return errors.New("wrong repository name format. Must be in the form owner/name, i.e(open-sauced/pizza)")
+	}
```
jpmcb, line 233: It'd be cool if we can _also_ accept `https://github.com/owner/repo` format and extract out the `owner` and `repo` since elsewhere in the CLI we accept that format (i.e., when using the `bake` command)

I hit this trying this with:
```
❯ pizza show https://github.com/open-sauced/api
Error: wrong repository name format. Must be in the form owner/name, i.e(open-sauced/pizza)
```

cmd/show/contributors.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// fetchUser: fetches all the user information (general info, and pull requests)
+func (m *ContributorModel) fetchUser() error {
+	var (
+		wg      sync.WaitGroup
+		errChan = make(chan error, 2)
+	)
+
+	wg.Add(2)
```
jpmcb, line 72: I usually prefer
```
wg.Add(1)
```
above each block that launches a go routine just so it's supremely clear where the waitgroup is getting added from. More a readability thing than anything.

jpmcb (commented):
Wow, this is great! 👏🏼 nice work on the dynamic side by side view

One thing I noticed is it does something weird when I resize my tmux window: the contributors disappear from the rendering. But then will re-appear if I keep resizing the window.

Screenshot 2023-09-05 at 3 24 53 PM

---

One sharp UI/UX edge I noticed is there's not much feedback as far as waiting for the next thing to happen. When I get into a contributor's view and hit <backspace>, it feels like it freezes since there's no spinning wheel or something (and I'm assuming it's blocking on the API call to return).

One thing to possibly do is explore caching some of the API results so you don't have to always fetch them. That could be a future enhancement too.

k1nho:
> One thing I noticed is it does something weird when I resize my tmux window: the contributors disappear from the rendering. But then will re-appear if I keep resizing the window.

Nice catch 🧤! just found out why this happened. Basically the window resizing message ```tea.WindowSizeMsg``` gets triggered on every window resize which is useful for responsiveness. However, if acted upon on ```Update``` rather than ```View``` by setting a particular width to the tables on that loop it creates the glitch effect. This should now be fixed on the latest commit.

> One sharp UI/UX edge I noticed is there's not much feedback as far as waiting for the next thing to happen. When I get into a contributor's view and hit , it feels like it freezes since there's no spinning wheel or something (and I'm assuming it's blocking on the API call to return).

Investigating this, need to review some elm architecture stuff with commands make sure is not the message flow. If anything, could probably throw a spinner on this one, but the cache would certainly be useful.

jpmcb:
👀

bdougie:
When testing this, noticing @dev-phantom's PRs are cut off and I am not sure if scrolling is built in. Repo is open-sauced/app

Screen Shot 2023-09-21 at 10 02 32 AM

k1nho:
Added responsiveness for contributor view in this last commit. This should now display horizontally if it does not fit in the vertical space. Also, introduced the list Model from bubbles instead of just the strings, I can see this format being more useful in the future. Now,  PR can be directly accessed with ```O``` with the browser dependency we are using for ```auth``` cmd as well.


https://github.com/open-sauced/pizza-cli/assets/59541661/2c61cf9e-e1a5-4797-98d4-ffcc8640b2f0

cmd/show/constants.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/charmbracelet/bubbles/list"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+)
+
+// WindowSize stores the size of the terminal
+var WindowSize tea.WindowSizeMsg
+
+// Keymaps
+var OpenPR = key.NewBinding(key.WithKeys("O"), key.WithHelp("O", "open pr"))
```
jpmcb, line 14: I ran into an interesting bug (we can fix this in a future patch if we want) where filtering on something that has a "O" opens the PR immediately. So a filter like:

```
/CODEOWNER
```
will open the first pr as soon as the first "O" is entered.

pkg/utils/github.go
```diff
@@ -2,27 +2,34 @@ package utils
...
-	// Remove the "https://github.com/" prefix from the URL
-	url = strings.TrimPrefix(url, "https://github.com/")
+// GetOwnerAndRepoFromURL: extracts the owner and repository name
+func GetOwnerAndRepoFromURL(input string) (owner, repo string, err error) {
+	var repoOwner, repoName string
 
-	// Split the remaining URL path into segments
-	segments := strings.Split(url, "/")
+	// check (https://github.com/owner/repo) format
+	u, err := url.Parse(input)
```
jpmcb, line 14: Nice. I'm definitly a fan of using `Parse` compared to the split / trim method.

pkg/utils/github.go
```diff
@@ -2,27 +2,34 @@ package utils
...
+	if err == nil && u.Host == "github.com" {
+		path := strings.Trim(u.Path, "/")
+		parts := strings.Split(path, "/")
+		if len(parts) != 2 {
+			return "", "", fmt.Errorf("Invalid URL: %s", input)
```
jpmcb, line 19: nit: Error strings shouldn't be capitalized

pkg/utils/github.go
```diff
@@ -2,27 +2,34 @@ package utils
...
-	// The first segment is the owner, and the second segment is the repository name
-	if len(segments) >= 2 {
-		owner = segments[0]
-		repo = segments[1]
-	} else {
-		return "", "", fmt.Errorf("invalid URL: %s", url)
+	// check (owner/repo) format
+	parts := strings.Split(input, "/")
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("Invalid URL: %s", input)
```
jpmcb, line 29: nit: Error strings shouldn't be capitalized

pkg/api/constants.go
```diff
@@ -0,0 +1,6 @@
+package api
+
+const (
+	APIEndpoint     = "https://api.opensauced.pizza"
+	BetaAPIEndpoint = "https://beta.api.opensauced.pizza"
+)
```
jpmcb, line 6: Looks like we have these constants here: https://github.com/open-sauced/pizza-cli/blob/f39986567f5d4e51b35d18bf4b2c1b81eddd09c0/pkg/constants/api.go#L1-L7

Can we use those ones?

jpmcb (approved):
Wow awesome work @k1nho - huge effort! 👏🏼 a few small nits if you want to fix otherwise we can get this in a roll forward fixes for those nits

bdougie:
> Wow awesome work @k1nho - huge effort! 👏🏼 a few small nits if you want to fix otherwise we can get this in a roll forward fixes for those nits

Agreed. This was a huge lift and getting this on beta would unblock a lot of future contributions and polish.

I am also featuring this briefly in my Redwoodjs conf lightning talk. It would be cool to get folks to use it after, but no rush.

cmd/show/constants.go
k1nho: Should now be fixed on the latest commit. On any ```tea.KeyMsg``` will now checks the internal ```filterState``` of the list to prevent the keymap collision.

jpmcb:
Closes https://github.com/open-sauced/pizza-cli/pull/24

Issue #28, opened 2023-08-18
Feature: OpenSauced API Golang client


jpmcb:
### Type of feature

🍕 Feature

### Current behavior

This probably needs to be in its own git repo / go module so it can be independently imported, but we should have some canotical way of generating and importing a Go client for the OpenSauced API.

This way, we will never have to worry about drift in our various Go projects.

### Suggested solution

We do generate a swagger doc and I believe there are ways to generate clients in different languages based on the swagger doc.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

bdougie:
Agreed. I am all for a separate repo (go-opensauced?) for a client. I think the open API spec will provide a path for it.

Anush008:
Just tried it out.
The output was impressive and offers several customization features.
https://openapi-generator.tech/docs/customization
Multiple installation options too.

Anush008:
Just tried it out.
The output was great. I found that there a quite some customisation options.
https://openapi-generator.tech/docs/templating

jpmcb:
Closed with the creation of https://github.com/open-sauced/go-api ✨

Issue #10, opened 2023-07-21
Feature: `pizza insights repo` for individual repo insights


jpmcb:
### Type of feature

🍕 Feature

### Current behavior

Once we introduce a `insights` subcommand, it may also be interesting to get individual repository insights.

### Suggested solution

A `repo` subcommand on the `insights` command that takes a github url and queries our API for information we've indexed would be interesting.

```
$ pizza insights repo <github-url>

ACTIVITY     PR OVERVIEW    PR VELOCITY    SPAM      CONTRIBUTORS
High         6 open PRs     30%            0         jpmcb, BekahHW, bdougie

See the full dashboard: https://insights.opensauced.pizza/{page-to-insights}
```

or with a `--output` flag:

```
$ pizza insights repo <github-url> --output json
{
  "activity": "high",
  "open-prs": 6,
  "pr-velocity": 0.3
  "spam": 0,
  "contributors": ["jpmcb", "BekahHW", "bdougie"]
}
```

This is mostly an idea but would be interesting to explore.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

cecobask:
@jpmcb, I will be happy to take on this issue!
It's quite similar to the work I did on #5.

cecobask:
Hi @jpmcb, I'm about to start working on this feature.
Should the command only take one repository at a time or I can extend it to accept multiple?

bdougie:
Agreed. The scope of the last PR increase a bit and would be better to do a follow up to make this work with a single repo for now and a follow up issue/pr for multiple repos.

PR #30, opened 2023-08-20
feat: provide repository contributors insights

### LINKED ISSUE #5: Feature: provide contributor insights
### Type of feature

🍕 Feature

### Current behavior

Currently, via the CLI, there's no way to see insights for a provided repos.

### Suggested solution

We should add an `insights` command that has a `contributors` subcommand to start.

When invoked with a git repo, it provides some very basic insights based on the commits to the user:

```
$ pizza insights contributors <git-repo>
Over last 30 days

New: 6
Recurring: 2
```

Should also provide the ability to specify the time range:
```
$ pizza insights contributors <git-repo> --time 60d

Over last 60 days:

New contributors: 6
Recurring contributors: 2
```

If the repo is a GitHub repo (which we have indexed data for from the API backend), we should generate a link to the insights page:

```
$ pizza insights contributors https://github.com/open-sauced/pizza-cli --time 60d

Over last 60 days:

New contributors: 6
Recurring contributors: 2

See more at: https://insights.opensauced.pizza/{insight-page}
```

The above will include a few calls to the API so there are some changes that probably need to be made there as well.

### Additional context

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

cecobask:
## Description
This pull request adds the feature to provide repository contributors insights.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Closes #5

### File contents:
```
$ cat repos.yaml

- https://github.com/kubernetes/kubernetes
- https://github.com/cli/cli
```

### Usage with `--file` / `-f` flag
```
$ pizza insights contributors --file repos.yaml

┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Repository URL      ┃ https://github.com/cli/cli ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ New contributors    ┃ 25                         ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Recent contributors ┃ 30                         ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Churn contributors  ┃ 40                         ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Repeat contributors ┃ 5                          ┃
┗━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Repository URL      ┃ https://github.com/kubernetes/kubernetes ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ New contributors    ┃ 158                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Recent contributors ┃ 280                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Churn contributors  ┃ 197                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Repeat contributors ┃ 122                                      ┃
┗━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

### Usage with `--repos` / `-r` flag
```
$ pizza insights contributors --repos "https://github.com/kubernetes/kubernetes,https://github.com/cli/cli"

┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Repository URL      ┃ https://github.com/cli/cli ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ New contributors    ┃ 25                         ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Recent contributors ┃ 30                         ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Churn contributors  ┃ 40                         ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Repeat contributors ┃ 5                          ┃
┗━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Repository URL      ┃ https://github.com/kubernetes/kubernetes ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ New contributors    ┃ 158                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Recent contributors ┃ 280                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Churn contributors  ┃ 197                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Repeat contributors ┃ 122                                      ┃
┗━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

### Ignore non-existent and non-indexed repositories
```
$ pizza insights contributors --repos "https://github.com/fake-owner/fake-repo,https://github.com/kubernetes/kubernetes"

ignoring repository issue: repository https://github.com/fake-owner/fake-repo is either non-existent or has not been indexed yet
┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Repository URL      ┃ https://github.com/kubernetes/kubernetes ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ New contributors    ┃ 158                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Recent contributors ┃ 280                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Churn contributors  ┃ 197                                      ┃
┣━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Repeat contributors ┃ 122                                      ┃
┗━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

### Error when neither `--file` or `--repos` flag was provided
```
$ pizza insights contributors

Error: must pass either --file or --repos flag
Usage:
  pizza insights contributors [flags]

Flags:
  -f, --file string     Path to yaml file containing an array of git repository urls
  -h, --help            help for contributors
  -p, --period int      Number of days, used for query filtering (default 30)
  -r, --repos strings   Comma-separated array of git repository urls

Global Flags:
      --beta              Shorthand for using the beta OpenSauced API endpoint ("https://beta.api.opensauced.pizza/v1"). supercedes the '--endpoint' flag
  -e, --endpoint string   The API endpoint to send requests to (default "https://api.opensauced.pizza/v1")

2023/08/20 19:36:16 must pass either --file or --repos flag
```

### Error when both `--file` and `--repos` flags were provided
```
$ pizza insights contributors --file repos.yaml --repos "https://github.com/kubernetes/kubernetes,https://github.com/cli/cli"

Error: if any flags in the group [file repos] are set none of the others can be; [file repos] were all set
Usage:
  pizza insights contributors [flags]

Flags:
  -f, --file string     Path to yaml file containing an array of git repository urls
  -h, --help            help for contributors
  -p, --period int      Number of days, used for query filtering (default 30)
  -r, --repos strings   Comma-separated array of git repository urls

Global Flags:
      --beta              Shorthand for using the beta OpenSauced API endpoint ("https://beta.api.opensauced.pizza/v1"). supercedes the '--endpoint' flag
  -e, --endpoint string   The API endpoint to send requests to (default "https://api.opensauced.pizza/v1")

2023/08/20 19:36:42 if any flags in the group [file repos] are set none of the others can be; [file repos] were all set
```

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed

## Added to documentation?

- [x] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook


jpmcb (changes requested):
I'm realizing this has quite abit of overlap with #24.

We'll want to align the user experience so that there is a clear story for CLI users with each command

This will also be blocked on us getting an API client for Go: https://github.com/open-sauced/pizza-cli/issues/28

pkg/api/constants.go
```diff
@@ -1,6 +1,7 @@
 package api
 
 const (
-	APIEndpoint     = "https://api.opensauced.pizza/v1"
-	BetaAPIEndpoint = "https://beta.api.opensauced.pizza/v1"
+	AlphaAPIEndpoint = "https://alpha.api.opensauced.pizza/v1"
```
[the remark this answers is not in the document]
brandonroberts: Not often. We used it before when prototyping against different infrastructure, but haven't needed to use it since then. I don't think we need to include it

cmd/insights/contributors.go
```diff
@@ -0,0 +1,309 @@
...
+		response, err := getRepositoryContributorInsights(opts.APIClient, "repeat", query)
+		if err != nil {
+			errorChan <- err
+			return
+		}
+		for _, data := range response.Data {
+			repoContributorInsights.Repeat = append(repoContributorInsights.Repeat, data.AuthorLogin)
+		}
+	}()
+	go func() {
```
k1nho, line 214: Very cool concurrency going on here! have been looking into also doing this on #24 as well to parallelize the data collection of the contributors so that the TUI can get the initial model started faster. Some questions if you don't mind. For this part, could it be the case that this is a dead goroutine when an error is returned from one of the data collectors, like the select is blocking once the error is acquired from ```errorChan``` closes the channel and then returns nil and err but waitGroup is not done, so ```doneChan``` does not get close?

Also,  could this be a good case to use [errGroup](https://pkg.go.dev/golang.org/x/sync@v0.3.0/errgroup)? @jpmcb

cecobask: Hi @k1nho, thanks!
The `waitGroup` will always close, regardless if an error occurred or not. This is because the start of each go routine explicitly defers the call to `waitGroup.Done()`.
In my opinion, this is a suitable place to use `errGroup`. It just comes down to style/preference. Behind the hood, they're doing something similar if I'm not wrong.

bdougie:
> I'm realizing this has quite abit of overlap with #24.
> 
> We'll want to align the user experience so that there is a clear story for CLI users with each command

I see the difference being the interactivity you get from the show. I didn't want to increase scope on #24 while we explore the basic interaction, but think we can do some more high level insights from commits and prs as well.

jpmcb:
@cecobask - the Go API client has landed ✨ https://github.com/open-sauced/go-api !!!! Are you able to give that a try? It likely has some sharp edges so please open issues in that repo as the need arises!

jpmcb:
Can you change this PR's target branch to be `beta`? 👀 We recently changed the default branch to `beta` in support of using the releaser action

For further reading:
- https://github.com/open-sauced/pizza-cli/issues/34

cecobask:
@jpmcb, Done 👍🏻 Another review session is in order 😄 
- Refactored the codebase to make API calls using the go client where possible, the only exception being `auth` and `repo-query` commands
- Consolidated duplicated logic and constants
- Configured `gci` linter with default settings
- Added concurrency to the `pizza bake` command
- Updated `pizza insights contributors` to use https://github.com/charmbracelet/bubbles#table for outputting results

Until https://github.com/open-sauced/go-api/issues/2 is resolved, `pizza bake` commands will fail with `400 Bad Request`.

cecobask:
The problems with [open-sauced/go-api](https://github.com/open-sauced/go-api/issues/2) were resolved.
I believe there are no more changes to be added here.
@jpmcb, can you please take a look when you get a chance?

main.go
```diff
@@ -12,11 +13,9 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-
 	utils.SetupRootCommand(rootCmd)
-
 	err = rootCmd.Execute()
 	if err != nil {
-		log.Fatal(err)
```
cecobask, line 20: Removed this log entry since the `Execute` method already outputs the error.
Previously, the error was outputted twice.

cmd/root/root.go
```diff
@@ -3,31 +3,50 @@ package root
...
+			betaFlag := cmd.Flags().Lookup(common.FlagNameBeta)
+			if betaFlag.Changed {
+				err := endpointFlag.Value.Set(api.BetaEndpointURL)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		},
+		SilenceUsage: true,
```
cecobask, line 42: Every time a command returned an error, the `help` output was also printed, together with the error.
I think this is cleaner. The help output will still be shown in case the user entered invalid input for commands or passed `--help` flag.

cmd/repo-query/repo-query.go
```diff
@@ -49,29 +46,23 @@ func NewRepoQueryCommand() *cobra.Command {
...
-			opts.APIClient = api.NewClient(endpoint)
-
+			opts.APIClient = api.NewClient(endpointURL)
 			opts.URL = args[0]
 			return run(opts)
 		},
 	}
-
 	cmd.Flags().StringVarP(&opts.branch, "branch", "b", "HEAD", "The branch to index")
-
```
jpmcb, line 74: nit: Spacing was for readability

cmd/repo-query/repo-query.go
```diff
@@ -85,7 +76,7 @@ func newRepoQueryAgent(apiClient *api.Client) *repoQueryAgent {
 	}
 }
 
-func (rq *repoQueryAgent) getOwnerAndRepo(url string) (owner, repo string, err error) {
+func GetOwnerAndRepoFromURL(url string) (owner, repo string, err error) {
```
jpmcb, line 79: Looks like it's getting used in `contributors.go` which is why it was changed from a method on the `repoQueryAgent` struct to a publicly facing package function.

This is a pretty general utility function, so we should abstract it out into some a utility package in `pkg/`. Maybe something like `pkg/github` since it's pretty specific to GitHub

cmd/root/root.go
```diff
@@ -4,30 +4,49 @@ package root
...
 	"github.com/open-sauced/pizza-cli/cmd/bake"
+	"github.com/open-sauced/pizza-cli/cmd/common"
+	"github.com/open-sauced/pizza-cli/cmd/insights"
 	repoquery "github.com/open-sauced/pizza-cli/cmd/repo-query"
 	"github.com/open-sauced/pizza-cli/pkg/api"
+	"github.com/spf13/cobra"
 )
 
+const ()
+
```
jpmcb, line 17: What's this empty constant block for?

cmd/root/root.go
jpmcb: > Every time a command returned an error, the help output was also printed

We should still dump the help message when an error is returned since it's usually due to some misconfiguration of flags or values. The standard approach in Cobra CLIs is to print the help message along with the error so the user can have a quick loop to correcting it.

pkg/api/constants.go
```diff
@@ -1,6 +1,46 @@
...
+	LocalhostEndpointKey = iota
+	ProdEndpointKey
+	BetaEndpointKey
+	AlphaEndpointKey
+	InvalidEndpointKey   = 666
+	LocalhostEndpointURL = "http://localhost:3001"
+	ProdEndpointURL      = "https://api.opensauced.pizza"
+	BetaEndpointURL      = "https://beta.api.opensauced.pizza"
+	AlphaEndpointURL     = "https://alpha.api.opensauced.pizza"
+	ToolsEndpointURL     = "https://opensauced.tools"
```
jpmcb, line 15: Thanks for adding these!! 👍🏼

pkg/api/constants.go
```diff
@@ -1,6 +1,46 @@
...
+	case ProdEndpointKey:
+		return ProdEndpointURL, nil
+	case BetaEndpointKey:
+		return BetaEndpointURL, nil
+	case AlphaEndpointKey:
+		return AlphaEndpointURL, nil
+	default:
+		return "", fmt.Errorf("unknown endpoint key %d", endpointKey)
+	}
+}
```
jpmcb, line 31: I'm generally not a fan of this approach. It always seems redundant. Why not just have callers get the endpoints directly through imports?

```go
myEndpoint := api.BetaEndpointURL
```

and I'm abit confused why we would need this? Generally, if a user provides `--endpoint`, we should take that whole sale and not check it against our keying of endpoints hard coded into the CLI. I may deploy my own API behind a local network and I want to provide that explicitly like:

```
pizza --endpoint https://api.opensauced.my-local-network.homelab ...
```

pkg/utils/root.go
```diff
@@ -2,11 +2,10 @@ package utils
 
 import "github.com/spf13/cobra"
 
-// SetupRootCommand is a convinence utility for applying templates and nice
+// SetupRootCommand is a convenience utility for applying templates and nice
 // user experience pieces to the root cobra command
 func SetupRootCommand(rootCmd *cobra.Command) {
 	cobra.AddTemplateFunc("wrappedFlagUsages", wrappedFlagUsages)
-
```
jpmcb, line 9: nit: This spacing was for readability

pkg/api/client.go
```diff
@@ -18,3 +24,15 @@ func NewClient(endpoint string) *Client {
...
+func DefaultClientSetup(cmd *cobra.Command) (context.Context, *client.APIClient, error) {
+	endpointURL := cmd.Flags().Lookup("endpoint").Value.String()
+	endpointKey, err := EndpointURLToKey(endpointURL)
+	if err != nil {
+		return nil, nil, err
+	}
+	configuration := client.NewConfiguration()
+	apiClient := client.NewAPIClient(configuration)
+	context := context.WithValue(context.Background(), client.ContextServerIndex, endpointKey)
+	return context, apiClient, nil
```
jpmcb, line 37: Why return the context? The background context should be handled by the API if for some reason it had a panic or exited early. And our control flow shouldn't be handling the background context anyways.

I wonder if it'd be better to do `context.TODO` since we're not really handling different context's control flow right now.

cmd/insights/contributors.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/spf13/cobra"
+)
+
+type Options struct {
+	// APIClient is the http client for making calls to the open-sauced api
+	APIClient *client.APIClient
+	// Repos is the array of git repository urls
+	Repos []string
+	// File is the path to yaml file containing an array of git repository urls
+	File string
```
jpmcb, line 26: nit: `FilePath` field makes more sense to me.

cmd/insights/contributors.go
```diff
@@ -0,0 +1,289 @@
...
+				return err
+			}
+			opts.APIClient = apiClient
+			return run(ctx, opts)
+		},
+	}
+	cmd.Flags().StringVarP(&opts.File, common.FlagNameFile, "f", "", "Path to yaml file containing an array of git repository urls")
+	cmd.Flags().StringSliceVarP(&opts.Repos, common.FlagNameRepos, "r", nil, "Comma-separated array of git repository urls")
+	cmd.Flags().Int32VarP(&opts.Period, common.FlagNamePeriod, "p", 30, "Number of days, used for query filtering")
+	cmd.MarkFlagsMutuallyExclusive(common.FlagNameFile, common.FlagNameRepos)
```
jpmcb, line 58: Same as in the bake command, I believe these flags being mutually exclusive is confusing. Instead, I would like to see:

```
pizza insights contributors URL
```
```
pizza insights contributors URL --file my-other-repos.yaml
```
```
pizza contributors insights --file my-repos.yaml
```

cmd/insights/contributors.go
```diff
@@ -0,0 +1,289 @@
...
+			if repoContributorInsights == nil {
+				return
+			}
+			repoContributorInsights.RenderTable()
+		}(repoURL)
+	}
+	go func() {
+		waitGroup.Wait()
+		close(doneChan)
+	}()
```
jpmcb, line 90: This looks like a similar structuring as the bake command refactor you have: I don't think having the `waitgroup.Wait()` in a separate thread make sense. The main through should just wait for these all to finish.

And similar comments to the `errorChan`: these will get squashed and only the first error will surface and be returned.

cmd/insights/contributors.go
```diff
@@ -0,0 +1,289 @@
...
+
+func (e *ErrRepoNotFoundOrIndexed) Error() string {
+	return fmt.Sprintf("repository %s is either non-existent or has not been indexed yet", e.RepoURL)
+}
+
+func NewErrRepoNotFoundOrIndexed(repoURL string) *ErrRepoNotFoundOrIndexed {
+	return &ErrRepoNotFoundOrIndexed{
+		RepoURL: repoURL,
+	}
+}
```
jpmcb, line 149: Having a structured error here seems unnecessary. Typically I've only seen these to be useful when different error types need to be compared across callers of APIs. Like when using `stat` and `os`:
```go
_, err := os.Stat(path)
if os.IsNotExist(err) { 
    // handle the file at path not existin
}
```
but it doesn't seem we're doing that here. Let's remove this struct type in favor of traditional stacked error handling in Go.

cmd/insights/contributors.go
```diff
@@ -0,0 +1,289 @@
...
+
+func findRepositoryByOwnerAndRepoName(ctx context.Context, apiClient *client.APIClient, repoURL string) (*client.DbRepo, error) {
+	owner, repoName, err := repoquery.GetOwnerAndRepoFromURL(repoURL)
```
jpmcb, line 152: `GetOwnerAndRepoFromURL` should be abstracted out of the `repoquery` package.

cmd/insights/contributors.go
```diff
@@ -0,0 +1,289 @@
...
+	go func() {
+		waitGroup.Wait()
+		close(doneChan)
+	}()
+	select {
+	case <-doneChan:
+		return repoContributorInsights, nil
+	case err = <-errorChan:
+		close(errorChan)
+		return nil, err
```
jpmcb, line 239: Same here: only the first error gets popped off the error channel. The others get squashed.

cmd/insights/contributors.go
```diff
@@ -0,0 +1,289 @@
...
+	repoContributorInsights := &repositoryContributorsInsights{
+		RepoID:  repo.Id,
+		RepoURL: repoURL,
+	}
+	var (
+		errorChan = make(chan error)
+		doneChan  = make(chan bool)
+		waitGroup = new(sync.WaitGroup)
+	)
+	waitGroup.Add(4)
```
jpmcb, line 185: I think I'd prefer if we explicitly added `1` each time for each `go func` block. Otherwies, if another gets added, it won't be clear to other contributors what's going on here.

jpmcb (changes requested):

pkg/api/constants.go
cecobask: For some reason, I assumed the go api client can only be used with the server configurations that it originally has.
I've refactored it to allow the use of any endpoint url.

cecobask (commented):
Thank you for the feedback, @jpmcb!
I've updated the code based on your suggestions.
Please, review at your convenience 👍🏻

jpmcb:
> Thank you for the feedback, @jpmcb!

Happy to provide the feedback! This is looking good, great work. Huge lift. This is on my docket to dive back into today!

pkg/utils/arguments.go
```diff
@@ -0,0 +1,29 @@
+package utils
+
+import (
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+func HandleRepositoryValues(repos []string, filePath string) (map[string]bool, error) {
+	uniqueRepoURLs := make(map[string]bool)
```
jpmcb, line 10: I think a `map[string]struct{}` may be slightly more efficient since it essentially creates an empty member on the right side which effectively implements a "set". But that's more a suggestion: take it or leave it!

cmd/insights/contributors.go
```diff
@@ -0,0 +1,271 @@
...
+type repositoryContributorsInsights struct {
+	RepoID  int32
+	RepoURL string
+	New     []string
+	Recent  []string
+	Alumni  []string
+	Repeat  []string
+}
+
+func (rci *repositoryContributorsInsights) RenderTable() {
```
jpmcb, line 100: Generally speaking, and this can be an additional issue we work on after this merges, we should also always provide a `-o / --output` flag that can be either "yaml" or "json" since we want people to be able to compose their data into other tools (like jq, grep, sed, awk, etc.)

jpmcb (commented):
This is getting very close!!!! Nice work!!!

cmd/insights/contributors.go
cecobask: I understand, I can refactor this to allow such an option as part of the work for https://github.com/open-sauced/pizza-cli/issues/10 👍🏻

cecobask:
> This is getting very close!!!! Nice work!!!

Thanks, @jpmcb! I've updated the branch with the latest changes 👍🏻 
Let me know if something else needs to be done here.

jpmcb (approved):
Great work!! Just one small nit on a regression I noticed in using a custom endpoint for the pizza service (which I typically do for testing local changes). This might be an issue with how the client generates it's endpoints. Otherwise, let's get this in! 💪🏼

cmd/bake/bake.go
```diff
@@ -39,127 +36,86 @@ type Options struct {
...
-
-			if useBeta {
-				fmt.Printf("Using beta API endpoint - %s\n", api.BetaAPIEndpoint)
-				endpoint = api.BetaAPIEndpoint
-			}
-			opts.APIClient = api.NewClient(endpoint)
-
-			opts.URLs = append(opts.URLs, args...)
+			endpointURL, _ := cmd.Flags().GetString(constants.FlagNameEndpoint)
+			opts.APIClient = api.NewGoClient(endpointURL)
```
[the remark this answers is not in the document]
cecobask: Hi @jpmcb, I think it works as expected.
I haven't set up the pizza service locally but the output I'm getting indicates that the go client is using the correct URL.
This is the error output I'm getting when I use a custom URL:
```
Error: error while calling 'PizzaOvenServiceAPI.BakeARepositoryWithThePizzaOvenMicroservice' with repository "https://github.com/jpmcb/gopherlogs": Post "http://localhost:8080/v1/bake": dial tcp [::1]:8080: connect: connection refused
```
I think you're getting a 404 error because the repository might not be indexed yet, but I might be wrong.

bdougie (approved):
I have been watching this quietly in the background and can say it works as expected for me. 

re: localhost. It might be worth setting up some follow up issues to identify things we can fix and enhance after this is merged.

Something top of mind is making these contributor insights actionable with some context on  % and story telling.

Perhaps we get @isabensusan to assist in adding this to the design backlog to assist in representing this in the UI.

cmd/bake/bake.go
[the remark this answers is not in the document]
jpmcb: Good insight. I believe this is because it's attempting to use a `v1` endpoint. So, when I provide `http://localhost:8080`, it's calling into `http://localhost:8080/v1/bake` which is the actual API endpoint, not the local pizza endpoint I have setup. It's doing the right thing since the client is reaching into the routes it knows about. This is good for now unless we have a different local dev story we want to explore 👍🏼

Issue #66, opened 2023-11-06
Bug: `--period` flag is broken


jpmcb:
### Describe the bug

The period flag seems to have broken at some point:

```
❯ pizza insights contributors https://github.com/open-sauced/app -p 45
Error: error while calling 'ContributorsServiceAPI.FindAllChurnPullRequestContributors' with repository 501028599': 400 Bad Request
error while calling 'ContributorsServiceAPI.NewPullRequestContributors' with repository 501028599': 400 Bad Request
error while calling 'ContributorsServiceAPI.FindAllRecentPullRequestContributors' with repository 501028599': 400 Bad Request
error while calling 'ContributorsServiceAPI.FindAllRepeatPullRequestContributors' with repository 501028599: 400 Bad Request

Usage:
  pizza insights contributors url... [flags]

Flags:
  -f, --file string    Path to yaml file containing an array of git repository urls
  -h, --help           help for contributors
  -p, --period int32   Number of days, used for query filtering (default 30)

Global Flags:
      --beta                Shorthand for using the beta OpenSauced API endpoint ("https://beta.api.opensauced.pizza"). Supersedes the
                            '--endpoint' flag
      --disable-telemetry   Disable sending telemetry data to OpenSauced
  -e, --endpoint string     The API endpoint to send requests to (default "https://api.opensauced.pizza")
  -o, --output string       The formatting for command output. One of: (table, yaml, csv, json) (default "table")
```

### Steps to reproduce

1. call the `insights` command with a repo and the `-p / --period` flag defined
2. Notice the 400, bad request responses

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Contributing Docs

- [X] I agree to follow this project's Contribution Docs

k1nho:
I think this might be because the API only accepts 7,30,90 as valid ```range``` values, so any other value just fails.

jpmcb:
Ahh, you're totally right. Great catch @k1nho - it's pretty confusing from the `pizza` CLI perspective to not have some of that information surfaced. Maybe we consider adding a client side restriction so you can only enter valid API values.

udaypatel1:
Just a note – this Issue seems like it should be closed as well, the related PR and change has been merged. (#67)

PR #67, opened 2023-11-09
fix: validate period flag

k1nho:
## Description
This PR adds a range validation to the ```insights contributors``` command. It will check for the value of the period flag. In case the period provided is invalid, it returns a descriptive error to signal that the API only accepts (7,30,90) as period values.

Before
```bash 
$ pizza insights contributors open-sauced/app -p 45
Error: error while calling 'ContributorsServiceAPI.FindAllRecentPullRequestContributors' with repository 501028599': 400 Bad Request
error while calling 'ContributorsServiceAPI.FindAllRepeatPullRequestContributors' with repository 501028599: 400 Bad Request
error while calling 'ContributorsServiceAPI.FindAllChurnPullRequestContributors' with repository 501028599': 400 Bad Request
error while calling 'ContributorsServiceAPI.NewPullRequestContributors' with repository 501028599': 400 Bad Request

Usage:
  pizza insights contributors url... [flags]
...
```

Now
```bash 
$ pizza insights contributors open-sauced/app -p 45
Error: invalid period: 45, accepts (7,30,90)

Usage:
  pizza insights contributors url... [flags]
...
```

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

#66 --period flag broken

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 docs.opensauced.pizza
- [ ] 🍕 dev.to/opensauced
- [ ] 📕 storybook

### DIFF
--- a/pkg/api/validation.go
+++ b/pkg/api/validation.go
@@ -1,0 +1,5 @@
+package api
+
+func IsValidRange(period int32) bool {
+	return period == 7 || period == 30 || period == 90
+}
--- a/cmd/insights/contributors.go
+++ b/cmd/insights/contributors.go
@@ -59,11 +59,15 @@ func NewContributorsCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVarP(&opts.FilePath, constants.FlagNameFile, "f", "", "Path to yaml file containing an array of git repository urls")
-	cmd.Flags().Int32VarP(&opts.Period, constants.FlagNamePeriod, "p", 30, "Number of days, used for query filtering")
+	cmd.Flags().Int32VarP(&opts.Period, constants.FlagNamePeriod, "p", 30, "Number of days, used for query filtering (7,30,90)")
 	return cmd
 }
 
 func (opts *contributorsOptions) run(ctx context.Context) error {
+	if !api.IsValidRange(opts.Period) {
+		return fmt.Errorf("invalid period: %d, accepts (7,30,90)", opts.Period)
+	}
+
 	repositories, err := utils.HandleRepositoryValues(opts.Repos, opts.FilePath)
 	if err != nil {
 		return err


pkg/api/validation.go
```diff
@@ -0,0 +1,8 @@
+package api
+
+func ValidateRange(period int32) bool {
+	if period == 7 || period == 30 || period == 90 {
+		return true
+	}
+	return false
```
nickytonline, line 7: refactor (non-blocking): No need for an explicit return of true or false unless this a preferred pattern in Go (I'm not a Gopher 😅)

```suggestion
	return period == 7 || period == 30 || period == 90
```

nickytonline (approved):
Looks great @k1nho! One non-blocking suggestion.

pkg/api/validation.go
k1nho: Not a pattern, might be a side effect from too much ```if err != nil``` 😂

jpmcb:
```go
return period == 7 || period == 30 || period == 90
```

is valid Go and it really comes down to personal preference: there will be people who say this is a prime use case for a `switch` in Go in order to reduce the boolean mind games someone reading your code might need to do to figure out what's going on. Like:

```go
switch range {
case 7:
	return true
case 30:
	return true
case 90:
	return true
default:
	// invalid range
	return false
}
```

but others would say that's too verbose and ends up being crufty 🤷🏼 

Anywho, I'm good with what's written here and makes sense to me!

pkg/api/validation.go
```diff
@@ -0,0 +1,5 @@
+package api
+
+func ValidateRange(period int32) bool {
```
jpmcb, line 3: Since this is a boolean function, can we rename this `IsValidRange`?

That typically reads better:

```go
if !api.IsValidRange(opts.Period)
```

jpmcb (approved):
Looks great! just one small suggestion!! Thanks for this - and thanks for your patience: I've been heads down on abunch of stuff the last few weeks!

pkg/api/validation.go
k1nho: thanks for the feedback!

pkg/api/validation.go
k1nho: > there will be people who say this is a prime use case for a switch in Go in order to reduce the boolean mind games

true, there is also this way with the switch.
```go
switch range {
case 7,30,90:
	return true
default:
	// invalid range
	return false
}
```
There is always the even-odd check meme way of doing it as well xD

jpmcb (approved):

PR #107, opened 2024-08-28
feat: now generate codeowners checks in user root but also repository root folder for .sauced.yaml

nickytonline:
## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [ ] Tier 4


pkg/config/config.go
```diff
@@ -34,7 +34,14 @@ func LoadConfig(path string) (*Spec, error) {
 
 	data, err := os.ReadFile(absPath)
 	if err != nil {
-		return nil, fmt.Errorf("error reading config file: %w", err)
+
+		// If the file does not exist, check if the fallback path exists
```
nickytonline, line 38: I'm not sure if what I'm doing is idiomatic Go @jpmcb. Also before it was returning the exact error, now I mention it failed reading file x or y. Should I still error out with the actual error like before?

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/posthog/posthog-go v1.2.19
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.9.0
```
nickytonline, line 16: This seems to be pretty popular in the Go ecosystem, but I defer to you @jpmcb for what you recommend for tests.

pkg/config/config_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLoadConfig(t *testing.T) {
```
nickytonline, line 11: For the changes I made, I thought it was out of scope in terms of testing to validate the contents of the YAML file, i.e. this but at the end of LoadConfig.

```go
err = yaml.Unmarshal(data, config)
if err != nil {
	return nil, fmt.Errorf("error unmarshaling config: %w", err)
}
```

pkg/config/config_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// Create an empty file for testing
+	configFilePath := filepath.Join(tmpDir, ".sauced.yaml")
+
+	if err := os.WriteFile(configFilePath, []byte{}, 0644); err != nil {
+		t.Fatalf("Failed to create empty file: %v", err)
+	}
+
+	// Path for a non-existent file
+	nonExistentPath := filepath.Join(tmpDir, "non_existent.yaml")
```
jpmcb, line 27: I'd move all of the setup above into the nested `t.Run(tt.name, func(t *testing.T) {`. That way, each test is completely independent and has all its own setup with its own temporary directory and temporary file.

Otherwise you run the risk of tests colliding or potentially impacting eachother if they were to mutate the file or dir in some way.

jpmcb (changes requested):

cmd/generate/codeowners/codeowners.go
```diff
@@ -72,7 +72,7 @@ func NewCodeownersCommand() *cobra.Command {
 			var err error
 
 			configPath, _ := cmd.Flags().GetString("config")
-			opts.config, err = config.LoadConfig(configPath)
+			opts.config, err = config.LoadConfig(configPath, opts.path+"/.sauced.yaml")
```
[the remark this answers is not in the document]
nickytonline: Makes sense. I would've done similar Node.

pkg/config/config_test.go
nickytonline: Yeah makes sense. I'd normally do that. I found an example and was just happy to get it running. 😅

go.mod
jpmcb, line 16: Can you run `go mod tidy` to update the `go.sum` file along side this new dependency?

pkg/config/config_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestLoadConfig(t *testing.T) {
+	var tmpDir string
+
+	setup := func() {
+		tmpDir = t.TempDir()
+	}
+
+	teardown := func() {
+		os.RemoveAll(tmpDir)
+	}
```
jpmcb, line 21: I'd recommend move all this down into individual test cases. In the future, we'll likely want to add `t.Parallel` to run parallel tests. And having any shared resource among tests (even a sole var like this) could cause problems.

Generally, I like to keep tests as flake free as possible with much their own resources.

jpmcb (approved):

go.mod
nickytonline: I ran `go mod tidy` but there were no changes. It looks like you added testify yesterday.

https://github.com/open-sauced/pizza-cli/blob/eb6946f7a4f61a97b5d953f13d83175d0bb0d9a9/go.sum#L123-L126

pkg/config/config_test.go
```diff
@@ -10,71 +10,45 @@ import (
...
+		nonExistentPath := filepath.Join(tmpDir, ".sauced.yaml")
+
+		config, err := LoadConfig(nonExistentPath, "")
+		assert.Error(t, err)
+		assert.Nil(t, config)
+	})
+
+	t.Run("Non-existent file with fallback", func(t *testing.T) {
+		t.Parallel()
+		tmpDir := t.TempDir()
```
nickytonline, line 38: Nice to know there's no teardown necessary for the temp files.

CleanShot 2024-08-28 at 22 08 08

brandonroberts (approved):

Issue #87, opened 2024-08-26
Feature: replace generated Go SDK library with hand rolled SDK


jpmcb:
### Suggested solution

There are alot of problems with the Go SDK that gets generated off our API's swagger doc: it's abit fragile and we haven't had bandwidth to build more robust testing and SDK service for clients like this CLI.

Instead, we should hand roll a small SDK in this codebase for the API endpoints we need.

This will also enable future work with the `pizza generate codeowners` in https://github.com/open-sauced/pizza-cli/issues/85

PR #110, opened 2024-08-29
chore: Remove the show command

jpmcb:
## Description

This removes the `show` command code entierly. While working on #87, there's alot of bits and pieces that would need to be refactored there. And this probably deserves its own refactor / re-write in the end.

And in all honesty, it probably doesn't make sense to compile an old piece of code that won't actually have any functionality for users (along with those dependencies).

For anyone looking at the old TUI code and possibly looking to exhume it, a good reference are the tags before 1.2.0: https://github.com/open-sauced/pizza-cli/releases/tag/v1.1.1-beta.5

Related to: #87 

## Steps to QA

1. Should be able to `just build` still

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


nickytonline (approved):

brandonroberts (approved):
🧹

PR #113, opened 2024-08-29
fix: mark endpoint and beta flags as hidden

### LINKED ISSUE #103: Bug: `--beta` and `--endpoint` should be secret, hidden flags
### Describe the bug

There's really no reason we need to ship and advertise the `--beta` or `--endpoint` flags: they should still be _valid_ but they should be hidden. They're most useful for super users or when doing dev on the CLI.

### Steps to reproduce

1. Run `pizza --help` and notice:

```
Global Flags:
      --beta                Shorthand for using the beta OpenSauced API endpoint ("https://beta.api.opensauced.pizza").
                            Supersedes the '--endpoint' flag
  -c, --config string       The saucectl config (default "~/.sauced.yaml")
      --disable-telemetry   Disable sending telemetry data to OpenSauced
  -e, --endpoint string     The API endpoint to send requests to (default "https://api.opensauced.pizza")
  -l, --log-level string    The logging level. Options: error, warn, info, debug (default "info")
  -o, --output string       The formatting for command output. One of: (table, yaml, csv, json) (default "table")
      --tty-disable         Disable log stylization. Suitable for CI/CD and automation
```

that both `--beta` and `--endpoint` are in the global flags

brandonroberts:
## Description

Closes #103 

## Steps to QA

1. Run `just build`
2. Run `./build/pizza -h`
3. Note the `beta` and `endpoint` flags are not shown

```
Usage:
  pizza generate codeowners path/to/repo [flags]

Flags:
  -h, --help                help for codeowners
      --owners-style-file   Whether to generate an agnostic OWNERS style file.
  -r, --range int           The number of days to lookback (default 90)

Global Flags:
  -c, --config string       The saucectl config (default "~/.sauced.yaml")
      --disable-telemetry   Disable sending telemetry data to OpenSauced
  -l, --log-level string    The logging level. Options: error, warn, info, debug (default "info")
  -o, --output string       The formatting for command output. One of: (table, yaml, csv, json) (default "table")
      --tty-disable         Disable log stylization. Suitable for CI/CD and automation
```

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -44,8 +44,16 @@ func NewRootCommand() (*cobra.Command, error) {
 	cmd.AddCommand(generate.NewGenerateCommand())
 	cmd.AddCommand(insights.NewInsightsCommand())
 	cmd.AddCommand(version.NewVersionCommand())
-	cmd.PersistentFlags().MarkHidden(constants.FlagNameEndpoint)
-	cmd.PersistentFlags().MarkHidden(constants.FlagNameBeta)
+
+	endpointHiddenErr := cmd.Flags().MarkHidden(constants.FlagNameEndpoint)
+	if endpointHiddenErr != nil {
+		return nil, endpointHiddenErr
+	}
+
+	betaHiddenErr := cmd.Flags().MarkHidden(constants.FlagNameBeta)
+	if betaHiddenErr != nil {
+		return nil, betaHiddenErr
+	}
 
 	return cmd, nil
 }
--- a/cmd/root/root.go
+++ b/cmd/root/root.go
@@ -44,6 +44,8 @@ func NewRootCommand() (*cobra.Command, error) {
 	cmd.AddCommand(generate.NewGenerateCommand())
 	cmd.AddCommand(insights.NewInsightsCommand())
 	cmd.AddCommand(version.NewVersionCommand())
+	cmd.PersistentFlags().MarkHidden(constants.FlagNameEndpoint)
+	cmd.PersistentFlags().MarkHidden(constants.FlagNameBeta)
 
 	return cmd, nil
 }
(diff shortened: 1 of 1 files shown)


cmd/root/root.go
```diff
@@ -45,6 +45,16 @@ func NewRootCommand() (*cobra.Command, error) {
...
+	errors := cmd.Flags().MarkHidden(constants.FlagNameEndpoint)
+	if errors != nil {
+		return nil, errors
+	}
+
+	errors = cmd.Flags().MarkHidden(constants.FlagNameBeta)
+	if errors != nil {
+		return nil, errors
+	}
+
```
jpmcb, line 57: Usually the convention is to call the error value that gets returned `err`. Calling it `errors` sort of makes it sound like there are multiple errors being returned when in reality it'd only ever be one error 

```suggestion
	err := cmd.Flags().MarkHidden(constants.FlagNameEndpoint)
	if err != nil {
		return nil, err
	}

	err = cmd.Flags().MarkHidden(constants.FlagNameBeta)
	if err != nil {
		return nil, err
	}

```

jpmcb (approved):

cmd/root/root.go
jpmcb: Just a note: there's also a pretty low likelihood that `cmd.Flags().MarkHidden` would even return an error. So this is the right way to handle it: just bubble it up.

I think the main case these would return an error is if `constants.FlagName...` wasn't defined in the array of command flags. And in that case, it'd be more a programmer error vs. a user error or system failure and you'd see it right away

nickytonline (approved):
:shipit:

PR #126, opened 2024-08-30
fix: Root command persistent flags are marked hidden correctly

jpmcb:
## Description

Root command now correctly marks persistent flags as hidden. It was giving:

```
❯ ./build/pizza
2024/08/30 11:01:05 error marking endpoint as hidden: flag "endpoint" does not exist
```

Followup to #113 

## Steps to QA

1. Build via `just build`
2. Run with `./build/pizza` and output is corrected

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4


nickytonline (approved):
Paired with @jpmcb on this so LGTM! 🚢

PR #135, opened 2024-09-04
fix: now the repo root .sauced.yaml gets it's contents read

nickytonline:
## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [ ] Tier 4


pkg/config/config.go
```diff
@@ -51,5 +59,7 @@ func LoadConfig(path string, repoRootPathConfig string) (*Spec, error) {
 		return nil, fmt.Errorf("error unmarshaling config: %w", err)
 	}
 
+	fmt.Printf("Loaded config %v", config)
```
jpmcb, line 62: nit: This is unnecessary and it'd be abit confusing to see the raw config output from the struct in the output - we can revisit more logging when we're closer to feature complete

pkg/config/config_test.go
```diff
@@ -35,14 +55,49 @@ func TestLoadConfig(t *testing.T) {
...
+
 		tmpDir := t.TempDir()
 		fallbackPath := filepath.Join(tmpDir, ".sauced.yaml")
-		require.NoError(t, os.WriteFile(fallbackPath, []byte("key: fallback"), 0644))
+		require.NoError(t, os.WriteFile(fallbackPath, []byte(fileContents), 0644))
+
+		// Print out the contents of the file we just wrote
+		writtenData, err := os.ReadFile(fallbackPath)
+		require.NoError(t, err)
+		fmt.Printf("Contents written to file:\n%s\n", string(writtenData))
```
jpmcb, line 81: nit: usually in Go, you won't see printing to stdout during tests: the test should either pass or error in an intelligent way that helps the developer know what broke

jpmcb (changes requested):

pkg/config/config.go
nickytonline: leftover from our debugging session. 😅

pkg/config/config_test.go
```diff
@@ -35,14 +55,48 @@ func TestLoadConfig(t *testing.T) {
...
+		// Print out the contents of the file we just wrote
+		_, err := os.ReadFile(fallbackPath)
+		require.NoError(t, err)
+
 		nonExistentPath := filepath.Join(tmpDir, "non-existent.yaml")
 
 		config, err := LoadConfig(nonExistentPath, fallbackPath)
+
+		fmt.Printf("Loaded config: %+v\n", config)
+		fmt.Printf("Error: %v\n", err)
```
jpmcb, line 87: nit: Same goes here - shouldn't log in the test at all

```suggestion
```

jpmcb (changes requested):

jpmcb (approved):

brandonroberts (approved):

PR #143, opened 2024-09-04
feat: now the documentation for the pizza-cli can be generated via pizza docs

### LINKED ISSUE #120: Feature: automate docs generation
### Suggested solution

Cobra offers several mechanisms for automatically generating markdown docs:

https://github.com/spf13/cobra/blob/main/doc/md_docs.go

Essentially, we would want a (maybe hidden?) command that we could use to generate docs:

```
pizza docs
```

This then would dump out the docs for the user. We could then use these docs in some automation to update either the docs in the GitHub or in a docs site.

nickytonline:
## Description

Now the documentation for the pizza-cli can be generated via the `docs` command, e.g. `pizza docs`. By default if no path is specified, the default path, `"./docs"` will be used.

Closes #120

## Steps to QA

1. Run `just build`
2. Run `./build/pizza docs`
3. Notice docs get generated in the `./docs` folder.
4. Run `./build pizza docs ./docs2`.
5. Notice docs get generated in the `./docs2` folder.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 # Ignore stuff from the builds that get generated.
 build/
+docs/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 # Ignore stuff from the builds that get generated.
 build/
-docs/
--- a/cmd/docs/docs.go
+++ b/cmd/docs/docs.go
@@ -24,7 +24,7 @@ func NewDocsCommand() *cobra.Command {
 		Short: "Generates the documentation for the CLI",
 		Args:  cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cmd.DisableAutoGenTag = true
+			cmd.Parent().Root().DisableAutoGenTag = true
 
 			// Use default path if no argument is provided
 			if len(args) == 0 {
--- a/cmd/docs/docs.go
+++ b/cmd/docs/docs.go
@@ -10,15 +10,49 @@ import (
 )
 
 type Options struct {
-	// the path to generate the documentation in
-	path string
+	Path string
 }
 
 const DefaultPath = "./docs"
 
-func NewDocsCommand() *cobra.Command {
-	opts := &Options{}
+func DeterminePath(args []string) (string, error) {
+	if len(args) == 0 {
+		fmt.Printf("No path was provided. Using default path: %s\n", DefaultPath)
+		return DefaultPath, nil
+	}
 
+	absPath, err := filepath.Abs(args[0])
+
+	if err != nil {
+		return "", err
+	}
+
+	return absPath, nil
+}
+
+func EnsureDirectoryExists(path string) error {
+	_, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		fmt.Printf("The directory %s does not exist. Creating it...\n", path)
+		return os.MkdirAll(path, os.ModePerm)
+	}
+	return err
+}
+
+func GenerateDocumentation(rootCmd *cobra.Command, path string) error {
+	fmt.Printf("Generating documentation in %s...\n", path)
+	err := doc.GenMarkdownTree(rootCmd, path)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Finished generating documentation in %s\n", path)
+
+	return nil
+}
+
+func NewDocsCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "docs",
 		Short: "Generates the documentation for the CLI",
@@ -26,40 +60,17 @@ func NewDocsCommand() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.Parent().Root().DisableAutoGenTag = true
 
-			// Use default path if no argument is provided
-			if len(args) == 0 {
-				opts.path = DefaultPath
-				fmt.Printf("No path was provided. Using default path: %s\n", DefaultPath)
-			} else {
-				absPath, err := filepath.Abs(args[0])
-				if err != nil {
-					return err
-				}
-				opts.path = absPath
-			}
+			path, err := DeterminePath(args)
 
-			// Create the directory if it doesn't exist
-			_, err := os.Stat(opts.path)
-			if os.IsNotExist(err) {
-				fmt.Printf("The directory %s does not exist. Creating it...\n", opts.path)
-
-				err := os.MkdirAll(opts.path, os.ModePerm)
-				if err != nil {
-					return fmt.Errorf("error creating documentation output directory %s: %s", opts.path, err)
-				}
-			}
-
-			// Generate markdown documentation
-			fmt.Printf("Generating documentation in %s...\n", opts.path)
-
-			err = doc.GenMarkdownTree(cmd.Parent().Root(), opts.path)
 			if err != nil {
 				return err
 			}
 
-			fmt.Printf("Finished generating documentation in %s\n", opts.path)
+			if err := EnsureDirectoryExists(path); err != nil {
+				return fmt.Errorf("error creating documentation output directory %s: %s", path, err)
+			}
 
-			return nil
+			return GenerateDocumentation(cmd.Parent().Root(), path)
 		},
 	}
 }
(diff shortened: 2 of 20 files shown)


cmd/docs/docs_test.go
```diff
@@ -0,0 +1,107 @@
...
+		// Check if a Markdown file was generated
+		files, err := os.ReadDir(DefaultPath)
+		assert.NoError(t, err)
+		assert.NotEmpty(t, files)
+
+		markdownFileFound := false
+
+		for _, file := range files {
+			if strings.HasSuffix(file.Name(), ".md") {
+				markdownFileFound = true
```
nickytonline, line 31: For the docs generation, I'm keeping it simple. I'm only checking if there was markdown generated as the text would change over time as the CLI evolves and I think mocking what `doc.GenMarkdownTree` does might be overkill.

cmd/docs/docs.go
```diff
@@ -0,0 +1,62 @@
...
+
+const DefaultPath = "./docs"
+
+func NewDocsCommand() *cobra.Command {
+	opts := &Options{}
+
+	return &cobra.Command{
+		Use:   "docs",
+		Short: "Generates the documentation for the CLI",
+		Args:  cobra.MaximumNArgs(1),
```
nickytonline, line 25: I'm not sure if this is best practices for Cobra.

cmd/docs/docs.go
```diff
@@ -0,0 +1,62 @@
...
+
+				err := os.MkdirAll(opts.path, os.ModePerm)
+				if err != nil {
+					return fmt.Errorf("error creating documentation output directory %s: %s", opts.path, err)
+				}
+			}
+
+			// Generate markdown documentation
+			fmt.Printf("Generating documentation in %s...\n", opts.path)
+			err = doc.GenMarkdownTree(cmd, opts.path)
```
jpmcb, line 52: Using this local cobra command will only generate docs for the `docs` command: you need to crawl back up the command tree to the root command to get docs generation for _all_ commands:  

```suggestion
			err = doc.GenMarkdownTree(cmd.Parent().Root(), opts.path)
```

Without this change:

```
❯ tree docs
docs
└── pizza_docs.md

1 directory, 1 file
```

With this change:

```
❯ tree docs
docs
├── pizza.md
├── pizza_completion.md
├── pizza_completion_bash.md
├── pizza_completion_fish.md
├── pizza_completion_powershell.md
├── pizza_completion_zsh.md
├── pizza_docs.md
├── pizza_generate.md
├── pizza_generate_codeowners.md
├── pizza_insights.md
├── pizza_insights_contributors.md
├── pizza_insights_repositories.md
├── pizza_insights_user-contributions.md
├── pizza_login.md
└── pizza_version.md

1 directory, 15 files
```

---

This suggested change would also break the tests you wrote since:

```go
		cmd := NewDocsCommand()
		err := cmd.Execute()
```

will panic since there's no parent or root for the command to crawl back up to.

In general, in my experience at least, using a `cobra.Command` directly in tests can get really really messy. If you have the full structure of the CLI (like with the root command) or things are well contained enough, it can work. But for this case, I'd say it's better to abstract out the things we want to test for (determining the file path) into its own function and test that.

cmd/root/root.go
```diff
@@ -43,6 +44,7 @@ func NewRootCommand() (*cobra.Command, error) {
 	cmd.AddCommand(generate.NewGenerateCommand())
 	cmd.AddCommand(insights.NewInsightsCommand())
 	cmd.AddCommand(version.NewVersionCommand())
+	cmd.AddCommand(docs.NewDocsCommand())
```
jpmcb, line 47: Mostly spit balling here: let's mark the docs command as hidden. It's really only a utility command for us to generate markdown docs to then drop in this repo.

And on that note, let's add the docs into this repository to enable @BekahHW to pick those up automatically into our docs site if we want:

```
❯ tree .
.
├── CHANGELOG.md
├── CODEOWNERS
├── Dockerfile
├── LICENSE
├── README.md
├── api
├── build
├── cmd
├── docs
│   ├── pizza.md
│   ├── pizza_completion.md
│   ├── pizza_completion_bash.md
│   ├── pizza_completion_fish.md
│   ├── pizza_completion_powershell.md
│   ├── pizza_completion_zsh.md
│   ├── pizza_docs.md
│   ├── pizza_generate.md
│   ├── pizza_generate_codeowners.md
│   ├── pizza_insights.md
│   ├── pizza_insights_contributors.md
│   ├── pizza_insights_repositories.md
│   ├── pizza_insights_user-contributions.md
│   ├── pizza_login.md
│   └── pizza_version.md
├── go.mod
├── go.sum
├── install.sh
├── justfile
├── main.go
├── npm
└── pkg

26 directories, 84 files
```

cmd/docs/docs.go
```diff
@@ -0,0 +1,62 @@
...
+const DefaultPath = "./docs"
+
+func NewDocsCommand() *cobra.Command {
+	opts := &Options{}
+
+	return &cobra.Command{
+		Use:   "docs",
+		Short: "Generates the documentation for the CLI",
+		Args:  cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
jpmcb, line 26: I'd go ahead and get rid of the `###### Auto generated by spf13/cobra on 5-Sep-2024` auto generated bit on the tail end of the markdown docs with:

```suggestion
		RunE: func(cmd *cobra.Command, args []string) error {
			cmd.DisableAutoGenTag = true

```

jpmcb (changes requested):

cmd/docs/docs.go
jpmcb: Yup, that's what those utility functions are for!! In a few places, I've defined our own function for `Args` that returns a `func(cmd *Command, args []string) error` and gives us more flexability:

https://github.com/open-sauced/pizza-cli/blob/1f6258c587979c4af5547e5123593417d1289705/cmd/generate/generate.go#L20

But either works great!

cmd/docs/docs.go
nickytonline: I was just coming here to mention that which is why I put it in draft mode. I figured I had to run it on the parent command. 😅

cmd/docs/docs.go
nickytonline: I set this via `cmd.Parent().Root().DisableAutoGenTag = true` so I didn't have to set it in individual commands.

nickytonline:
Just need to update tests.

cmd/docs/docs.go
nickytonline: For the testing, I was definitely not trying to test GenMarkdownTree, just trying to do a sanity check that markdown was generated, but given it runs at the root command now, I definitely agree it could get hairy since I'd probably have to mock a root command etc.

I've broken it up now to test folder creation, and like before, the assumption is GenMarkdownTree works (tested in Cobra I imagine).

.gitignore
```diff
@@ -1,2 +1,3 @@
 # Ignore stuff from the builds that get generated.
 build/
+docs/
```
jpmcb, line 3: We _should_ commit the `docs/` markdown to the repo so we can automatically generate it and integrate it

cmd/docs/docs.go
```diff
@@ -0,0 +1,76 @@
...
+		return DefaultPath, nil
+	}
+
+	absPath, err := filepath.Abs(args[0])
+
+	if err != nil {
+		return "", err
+	}
+
+	return absPath, nil
```
jpmcb, line 30: This chunk can be simplified to:

```suggestion
	return filepath.Abs(args[0])
```

jpmcb (changes requested):

.gitignore
nickytonline: Isn't it more of a build artifact? Every PR has the potential to make the docs out of date.

That said, we could have the pizza-action generate docs whenever a PR goes up.

nickytonline: Anyway, that said, I've added the first iteration of the CLI docs to the PR.

jpmcb: > Isn't it more of a build artifact

Yes, it's essentially a build artifact. But we also need living docs in this repository. And, ultimately, `pizza docs` should serve the purpose of making it as easy and seamless for us to keep those up to date with the code.

I'd imagine the next step here is to have another step on the release action that runs `pizza docs`, updates the docs under the `docs/` directory in the repo, and then commits it.

jpmcb (approved):

brandonroberts (approved):

Issue #132, opened 2024-09-03
Bug: Testing - `TestLoadConfig` tests for `Default_pat` break if running on system with a `.sauced.yaml`


jpmcb:
### Describe the bug

The `TestLoadConfig/Default_path` test breaks if run on a system that already has a `~/.sauced.yaml` config. For example, on my system: 

```
--- FAIL: TestLoadConfig (0.00s)
    --- FAIL: TestLoadConfig/Default_path (0.00s)
        config_test.go:51:
                Error Trace:    /Users/jpmcb/workspace/opensauced/pizza-cli/pkg/config/config_test.go:51
                Error:          An error is expected but got nil.
                Test:           TestLoadConfig/Default_path
        config_test.go:52:
                Error Trace:    /Users/jpmcb/workspace/opensauced/pizza-cli/pkg/config/config_test.go:52
                Error:          Expected nil, but got: &config.Spec{Attributions:map[string][]string{"bdougie":[]string{"brian@opensauced.pizza"}, "brandonroberts":[]string{"brandon@opensauced.pizza"}, "jpmcb":[]string{"jpmmcbride@gmail.com", "john@opensauced.pizza"}, "nickytonline":[]string{"nick@nickyt.co"}, "zeucapu":[]string{"coding@zeu.dev"}}}
                Test:           TestLoadConfig/Default_path
```

Note that it expects to find an empty config but instead finds my config in my home directory.

Ideally, any tests that run against the filesystem should have some sort of abstract filesystem paradigm so that there are no flakes if there is an existing environment with existing artifacts.

### Steps to reproduce

1. Create a `~/.sauced.yaml` - populate it with some yaml confings
2. Run the tests with `just test`
3. Notice test fails

PR #141, opened 2024-09-04
feat: Create a contributor list after generating codeowners

jpmcb:
## Description

Implements a loop-back user flow for connecting discovered codeowners with the OpenSauced platform. This will ask if the user wants to create a Contributor Insight and makes the necessary API calls to make that happen.

## Related Tickets & Documents

Closes #https://github.com/open-sauced/pizza-cli/issues/99

## Mobile & Desktop Screenshots/Recordings

Example workflow:

```
❯ ./build/pizza generate codeowners . -l debug
Built logger with log level: 4
Opened repo at: /Users/jpmcb/workspace/opensauced/pizza-cli
Looking back 90 days
Finished processing commits for: /Users/jpmcb/workspace/opensauced/pizza-cli
Processing codeowners file at: /Users/jpmcb/workspace/opensauced/pizza-cli/CODEOWNERS
Finished generating file: /Users/jpmcb/workspace/opensauced/pizza-cli/CODEOWNERS
Do you want to add these codeowners to an OpenSauced Contributor Insight? (y/n): y
Adding codeowners to contributor insight
Initiating log in flow
Looking up OpenSauced workspace: Pizza CLI
Query user workspaces page: 1
Found existing workspace named: Pizza CLI
Found workspace: Pizza CLI
Looking up Contributor Insight for local repository: pizza-cli
Query user Workspace Contributor Insight page: 1
Found existing Workspace Contributor Insight named: pizza-cli
Updating Contributor Insight with codeowners with GitHub aliases: [zeucapu jpmcb nickytonline]
Updated Contributor Insight for local repository: pizza-cli
Access list on OpenSauced:
https://app.opensauced.pizza/workspaces/8718805b-66b8-4da5-887d-c37366b8e0da/contributor-insights/dfa8e053-f68a-47e9-ae09-fc13579ca4a5
```

(Note: this is the verbos logging but not where it asks `Do you want to add these codeowners to an OpenSauced Contributor Insight?` and I respond with `y`)

## Steps to QA

1. Pull this branch and build a binary with `just build`
2. Execute the command via `pizza generate codeowners`

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

PR #146, opened 2024-09-05
chore: Remove flaky default config test

jpmcb:
## Description

This removes a flaky test that wasn't doing much for us: in the case that there's a valid `~/.sauced.yaml` in the user's home directory, this test would blow up. 

We _could_ in theory build some tooling to completely isolate the tests in a separate in memory file system, but that would be alot of effort for standard lib functionality that would probably be better served by e2e tests.

Closes: https://github.com/open-sauced/pizza-cli/issues/132

## Steps to QA

1. Run `just test` - tests pass with a valid `~/.sauced.yaml` in the user's home directory:

```
❯ file ~/.sauced.yaml
/Users/jpmcb/.sauced.yaml: ASCII text
```

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -94,11 +94,4 @@ attribution:
 		assert.Equal(t, []string{"nick@nickyt.co", "nick@opensauced.pizza"}, config.Attributions["nickytonline"])
 		assert.Equal(t, []string{"coding@zeu.dev"}, config.Attributions["zeucapua"])
 	})
-
-	//t.Run("Default path", func(t *testing.T) {
-	//t.Parallel()
-	//config, err := LoadConfig(DefaultConfigPath, "")
-	//assert.Error(t, err)
-	//assert.Nil(t, config)
-	//})
 }


pkg/config/config_test.go
```diff
@@ -94,11 +94,4 @@ attribution:
 		assert.Equal(t, []string{"nick@nickyt.co", "nick@opensauced.pizza"}, config.Attributions["nickytonline"])
 		assert.Equal(t, []string{"coding@zeu.dev"}, config.Attributions["zeucapua"])
 	})
-
```
nickytonline, line 97: Did you want to delete this instead of comment it out?

nickytonline (approved):
🚢

pkg/config/config_test.go
jpmcb: I had commented it out and committed it in https://github.com/open-sauced/pizza-cli/pull/141 since tests would fail with this: this just gets rid of it. I played around abit with getting [`spf13/afero`'s](https://github.com/spf13/afero) abstract, in memory filesystem to work with this. But it'd be alot of tooling / boilerplate / abstraction for a test that basically just makes sure the stdlib is doing what we expect.

Still probably worthwhile in an end to end test but that'd also probably have to be in a container or something.

Anyways, worth punting on 🏈

brandonroberts (approved):

Issue #125, opened 2024-08-30
Feature: setup and validate Posthog


jpmcb:
We already have Posthog integrated into the CLI:

https://github.com/open-sauced/pizza-cli/blob/ad9e3a0dd2ade42dbfdd1f6010af46eda35efb30/pkg/utils/posthog.go#L12-L16

This gets built with our public token. We'll want to make sure we're logging relevant data in posthog for codeowners generation of all kinds.

---

- [ ] Ensure build pipelines all have the public token as part of their build pipeline
- [ ] Ensure dashboards on our internal Posthog are setup

Issue #119, opened 2024-08-30
Feature: dev-ex - `just help`


jpmcb:
### Suggested solution

There are a few things we could do with `just` to make it abit easier to use:

- include a `just help` command to make the onboarding experience abit easier
- document each command

nickytonline:
This should be all we need to do along with just documenting commands that aren't already documented.

```
help:
  @echo "Available commands:"
  @just --list
```

PR #147, opened 2024-09-05
fix: Revive Posthog's write only key

jpmcb:
## Description

This PR does the following:
- Exhumes the posthog reporting including the buildtime variable. The justfile loads the posthog public API key from the new `.env`
  - closes: https://github.com/open-sauced/pizza-cli/issues/125
  - Adds posthog capturing for codeowners generate commands and insight commands
- Introduces a `just help` command to get a list of all commands (and their descriptions)
  - closes: https://github.com/open-sauced/pizza-cli/issues/119
- Upgrades golangci-lint to `golangci-lint:v1.60`

## Steps to QA

1. Build locally
2. For the core engineering team, you can log in and see the events being captured by the CLI

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pkg/utils/posthog.go
+++ b/pkg/utils/posthog.go
@@ -12,12 +12,13 @@ var (
 // PosthogCliClient is a wrapper around the posthog-go client and is used as a
 // API entrypoint for sending OpenSauced telemetry data for CLI commands
 type PosthogCliClient struct {
-	client posthog.Client
+	client    posthog.Client
+	activated bool
 }
 
 // NewPosthogCliClient returns a PosthogCliClient which can be used to capture
 // telemetry events for CLI users
-func NewPosthogCliClient() *PosthogCliClient {
+func NewPosthogCliClient(activated bool) *PosthogCliClient {
 	client, err := posthog.NewWithConfig(
 		writeOnlyPublicPosthogKey,
 		posthog.Config{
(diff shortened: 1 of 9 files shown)


brandonroberts (approved):

pkg/utils/posthog.go
```diff
@@ -44,44 +46,106 @@ func (p *PosthogCliClient) Done() {
...
-	})
+	if p.activated {
+		p.client.Enqueue(posthog.Capture{
+			DistinctId: "login-failures",
+			Event:      "cli_user failed log in",
+		})
+	}
+}
+
+//nolint:errcheck
```
nickytonline, line 73: Are the `nolint:errcheck`s here because Posthog is fire and forget?

nickytonline (approved):
:ship:

pkg/utils/posthog.go
jpmcb: Right: Posthog's client interface for the Enqueue function returns an error:

```
	// Queues a message to be sent by the client when the conditions for a batch
	// upload are met.
	// This is the main method you'll be using, a typical flow would look like
	// this:
	//
	//	client := posthog.New(apiKey)
	//	...
	//	client.Enqueue(posthog.Capture{ ... })
	//	...
	//	client.Close()
	//
	// The method returns an error if the message queue could not be queued, which
	// happens if the client was already closed at the time the method was
	// called or if the message was malformed.
	Enqueue(Message) error
```

And we don't really care (at least at this point) to log to the user that telemetry failed.

PR #157, opened 2024-09-06
chore: now docs are autogenerated as part of the release process

### LINKED ISSUE #152: Feature: automate docs generation in release action
### Suggested solution

Now that #143 has landed to enable the `pizza docs` command, we can use this in the release action to automate generating and committing new docs during the release action.

In https://github.com/open-sauced/pizza-cli/pull/151, I included the `just gen-docs` command which will execute `go run main.go docs` to generate the docs. We can use this with the existing `just` and `go` setup to _also_ generate and commit the docs during the release action here:

https://github.com/open-sauced/pizza-cli/blob/3cdf5ceed8d0a273a976fdcd6e0ed94e94cab265/.github/workflows/release.yaml#L80-L86

nickytonline:
## Description

This pull request introduces a new job to the release workflow for automating the documentation update process before releasing a new version. Additionally, it ensures that the release process waits for the documentation update PR to be merged before proceeding.

The repository is already configured for GitHub Actions to create and approve PRs.

CleanShot 2024-09-06 at 14 43 34

Closes #152

## Steps to QA

I've tested my script, but not the whole flow. We can see how it behaves for a beta release.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,7 +16,55 @@ jobs:
     name: Test, lint, & build
     uses: ./.github/workflows/test.yaml
 
+  prerelease:
+    name: Update documentation
+    runs-on: ubuntu-latest
+    outputs:
+      pr_number: ${{ steps.create_pr.outputs.pr_number }}
+    steps:
+      - name: Generate token
+        id: generate_token
+        uses: tibdex/github-app-token@v2
+        with:
+          app_id: ${{ secrets.OS_GITHUB_APP_ID }}
+          private_key: ${{ secrets.OS_GITHUB_APP_PRIVATE_KEY }}
+
+      - name: "☁️ checkout repository"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ steps.generate_token.outputs.token }}
+
+      - name: "🤲 Setup Just"
+        uses: extractions/setup-just@v2
+
+      - name: "📗 generate documentation"
+        run: just gen-docs
+
+      - name: Create PR for documentation update
+        id: create_pr
+        env:
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+        run: |
+          timestamp=$(date +"%Y%m%d%H%M%S")
+          branch_name="chore-automated-doc-updates-$timestamp"
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git checkout -b "$branch_name"
+          git add .
+          if git diff --staged --quiet; then
+            echo "No changes to commit"
+            echo "pr_number=" >> $GITHUB_OUTPUT
+          else
+            git commit -m "chore: update documentation ($timestamp)"
+            git push -u origin "$branch_name"
+            pr_number=$(gh pr create --title "chore: update documentation ($timestamp)" --body "Automated documentation update for release" --base ${{ github.ref_name }} --head "$branch_name")
+            echo "pr_number=$pr_number" >> $GITHUB_OUTPUT
+            gh pr merge $pr_number --auto --delete-branch --squash
+          fi
+
   release:
+    needs: [test, prerelease]
     outputs:
         release-tag: ${{ steps.semantic-release.outputs.release-tag }}
     name: Semantic release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -16,7 +16,55 @@ jobs:
...
+          else
+            git commit -m "chore: update documentation ($timestamp)"
+            git push -u origin "$branch_name"
+            pr_number=$(gh pr create --title "chore: update documentation ($timestamp)" --body "Automated documentation update for release" --base ${{ github.ref_name }} --head "$branch_name")
+            echo "pr_number=$pr_number" >> $GITHUB_OUTPUT
+            gh pr merge $pr_number --auto --delete-branch --squash
+          fi
+
   release:
+    needs: [test, prerelease]
```
nickytonline, line 67: This seems like a good approach, but would love feedback. Adding it as a step without a `needs` means the PR might not be merged for the docs yet before a release happens, and it can also cause an infinite loop of a release.

I put tests in here as well.

.github/workflows/release.yaml
```diff
@@ -46,6 +94,25 @@ jobs:
...
+            pr_state=$(gh pr view $pr_number --json state -q .state)
+            if [ "$pr_state" = "MERGED" ]; then
+              echo "PR has been merged"
+              break
+            elif [ "$pr_state" = "CLOSED" ]; then
+              echo "PR was closed without merging"
+              exit 1
+            fi
+            echo "Waiting for PR to be merged..."
+            sleep 10
```
nickytonline, line 113: There is a risk of this never finishing, so we could add a retry max in here.

.github/workflows/release.yaml
```diff
@@ -46,6 +94,25 @@ jobs:
       - name: "🔧 install npm@latest"
         run: npm i -g npm@latest
 
+      - name: "Wait for PR to be merged"
+        if: needs.prerelease.outputs.pr_number != ''
```
nickytonline, line 98: We skip the `"Wait for PR to be merged"` if no PR was created, i.e. up top

```
if git diff --staged --quiet; then
  echo "No changes to commit"
  echo "pr_number=" >> $GITHUB_OUTPUT
```

.github/workflows/release.yaml
```diff
@@ -16,7 +16,55 @@ jobs:
...
+
+      - name: "📗 generate documentation"
+        run: just gen-docs
+
+      - name: Create PR for documentation update
+        id: create_pr
+        env:
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+        run: |
+          timestamp=$(date +"%Y%m%d%H%M%S")
```
nickytonline, line 49: These could be a script but I opted for inlining for simplicity.

.github/workflows/release.yaml
jpmcb: I'd honestly go even simpler: we already automatically commit the `CHANGELOG.md` in the release action. So I'd say this could just be an additional step before the semantic-release flow (although, I think semantic-release fails if the git repo is dirty so it'd probably have to be a separate, quick followup commit).

So instead of creating a whole PR, I'd just commit it directly to the target release branch with `skip-ci` like we do with other release assets.

Although I could see the value in getting eyes on the generated docs via PR before merging them in. It could be easy to let the process get away from us and we never actually look at what's being generated.

.github/workflows/release.yaml
```diff
@@ -16,7 +16,55 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ steps.generate_token.outputs.token }}
+
+      - name: "🤲 Setup Just"
+        uses: extractions/setup-just@v2
+
+      - name: "📗 generate documentation"
+        run: just gen-docs
```
jpmcb, line 42: This is missing a "setup go" step:

```yaml
    - name: "🐹 Setup Go"
      uses: actions/setup-go@v5
      with:
        go-version: 1.22.x
```

I don't _think_ the GitHub runners have Go installed, but before running `just gen-docs` (which executes `go run`), I'd go ahead and get Go setup

.github/workflows/release.yaml
nickytonline: I wasn't sure if we wanted to commit and push, so sounds good!

jpmcb (approved):

brandonroberts (approved):

PR #158, opened 2024-09-06
fix: use the local directory and home directory as fallback for .sauced.yaml

brandonroberts:
## Description

Uses the local `.sauced.yaml` as the default, and `~/.sauced.yaml` as the fallback location for finding the configuration file

## Steps to QA

1. Run just build
2. Go to a repo with no `.sauced.yaml` file in its root, but with a `~/.sauced.yaml` file
3. Run the codeowners generation command `pizza generate codeowners .`
4. Go to a repo with a `.sauced.yaml` file in its root
5. Run the codeowners generation command `pizza generate codeowners .`
6. Go to a repo with a `.sauced.yaml` file in its root
7. Run the codeowners generation command with a custom location `pizza generate codeowners . -c ../../.sauced.yaml`

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [x] Tier 4

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -30,7 +30,7 @@ attribution:
 
 		require.NoError(t, os.WriteFile(configFilePath, []byte(fileContents), 0600))
 
-		config, err := LoadConfig(configFilePath, "")
+		config, err := LoadConfig(configFilePath)
 		require.NoError(t, err)
 		assert.NotNil(t, config)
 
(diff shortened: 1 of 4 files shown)


pkg/config/config_test.go
```diff
@@ -30,7 +30,7 @@ attribution:
 
 		require.NoError(t, os.WriteFile(configFilePath, []byte(fileContents), 0600))
```
nickytonline, line 31: We should add a test when `./sauced.yaml` doesn't exist, but `~/.sauced.yaml` does.

jpmcb: > We should add a test when ./sauced.yaml doesn't exist, but ~/.sauced.yaml does.

This is essentially the test I removed in: https://github.com/open-sauced/pizza-cli/pull/146 - in the case that this runs and the developer doesn't have a `~/.sauced.yaml`, a simple test would fail. If we attempt to stub one for the user, this could conflict with people who _do_ have a `~/.sauced.yaml` - there's alot of tricky scaffolding and edge cases for people running these tests with or without a home dir `~/.sauced.yaml`. Truly isolating this test probably needs to run in a container as e2e suites.

In the future, might be worth looking at Dagger since their CI/CD approach would solve this use case (running tests / builds / etc. in isolated environments within containers that can pass artifacts to eachother)

TLDR: we should punt on more complex testing that requires abunch of scaffolding

jpmcb (approved):

pkg/config/config_test.go
nickytonline: Sounds good. Forgot about that. Thanks for the reminder.

nickytonline (approved):
:shipit:

PR #223, opened 2024-10-14
feat: `--output-path` flag for `pizza generate codeowners`

### LINKED ISSUE #222: Feature: Specify Output Directory for CODEOWNERS file
### Suggested solution

In it's current state, the following are available options for generating a CODEOWNERS file via the CLI:

```
Flags:
  -h, --help                help for codeowners
      --owners-style-file   Generate an agnostic OWNERS style file instead of CODEOWNERS.
  -r, --range int           The number of days to analyze commit history (default 90) (default 90)

Global Flags:
  -c, --config string       The codeowners config
      --disable-telemetry   Disable sending telemetry data to OpenSauced
  -l, --log-level string    The logging level. Options: error, warn, info, debug (default "info")
      --tty-disable         Disable log stylization. Suitable for CI/CD and automation
```

I want to specify an output directory whereby I can put my generated CODEOWNERS file.

Example: `pizza generate codeowners . -o .bitbucket/`

In Bitbucket SCM, the CODEOWNERS file must be under `.bitbucket/` and so this way, I can explicitly define where to put the auto-generated file.

This would natively support an output flag `-o` for this portion of the CLI and give flexibility in defining where to put the CODEOWNERS file in your repository.

Thanks! 😁🚀

udaypatel1:
## Description

This PR adds logic to handle an optional output directory flag value to the `generate codeowners` command

Mirrors the expected behavior of the same flag that exists for `generate config`

Resolves #222

I am not sure how to self-assign the ticket – if an admin could assign the Issue to me, that'd be wonderful :)

output

## Steps to QA

1. Build the CLI
2. Run `pizza generate codeowners` with and without the `-o` / `--output-path` flag

* Testing Permutations
    1. `pizza generate codeowners .`
        - should modify or create a CODEOWNERS file
    2. `pizza generate codeowners . -o /path/to/some/directory/`
        - should modify or create a CODEOWNERS file in the directory path provided
    4. `pizza generate codeowners . -o /path/to/some/directory/ --owners-style-file`
        - should modify or create an OWNERS file in the directory path provided
        
All other functionalities should remain unchanged and BAU. The optional flag should just create or modify in a specified file path that is given.

## Tier (staff will fill in)

- [ ] Tier 1
- [ ] Tier 2
- [ ] Tier 3
- [ ] Tier 4

## [optional] What gif best describes this PR or how it makes you feel?

happy-feet

Please let me know if I'm missing anything or there are complications! Thank you! 😁

### DIFF
--- a/cmd/generate/codeowners/codeowners.go
+++ b/cmd/generate/codeowners/codeowners.go
@@ -26,6 +26,9 @@ type Options struct {
 	// The default should be to generate a GitHub style "CODEOWNERS" file.
 	ownersStyleFile bool
 
+	// where the output file will go
+	outputPath string
+
 	// the number of days to look back
 	previousDays int
 
@@ -74,6 +77,9 @@ pizza generate codeowners . --owners-style-file
 
 # Specify a custom location for the .sauced.yaml file
 pizza generate codeowners . --config /path/to/.sauced.yaml
+
+# Specify a custom output location for the CODEOWNERS file
+pizza generate codeowners . --output-path /path/to/directory
 		`,
 		Args: func(_ *cobra.Command, args []string) error {
 			if len(args) != 1 {
@@ -113,6 +119,13 @@ pizza generate codeowners . --config /path/to/.sauced.yaml
 			}
 
 			opts.ownersStyleFile, _ = cmd.Flags().GetBool("owners-style-file")
+			opts.outputPath, _ = cmd.Flags().GetString("output-path")
+
+			// Default the outputPath to the base path if no flag value is given
+			if opts.outputPath == "" {
+				opts.outputPath = opts.path
+			}
+
 			opts.previousDays, _ = cmd.Flags().GetInt("range")
 			opts.tty, _ = cmd.Flags().GetBool("tty-disable")
 
@@ -139,6 +152,7 @@ pizza generate codeowners . --config /path/to/.sauced.yaml
 
 	cmd.PersistentFlags().IntP("range", "r", 90, "The number of days to analyze commit history (default 90)")
 	cmd.PersistentFlags().Bool("owners-style-file", false, "Generate an agnostic OWNERS style file instead of CODEOWNERS.")
+	cmd.PersistentFlags().StringP("output-path", "o", "", "Directory to create the output file.")
 
 	return cmd
 }
@@ -176,21 +190,23 @@ func run(opts *Options, cmd *cobra.Command) error {
 		return fmt.Errorf("error traversing git log: %w", err)
 	}
 
-	// Bootstrap codeowners
-	var outputPath string
+	// Define which file to generate based on a flag
+	var fileType string
 	if opts.ownersStyleFile {
-		outputPath = filepath.Join(opts.path, "OWNERS")
+		fileType = "OWNERS"
 	} else {
-		outputPath = filepath.Join(opts.path, "CODEOWNERS")
+		fileType = "CODEOWNERS"
 	}
 
-	opts.logger.V(logging.LogDebug).Style(0, colors.FgBlue).Infof("Processing codeowners file at: %s\n", outputPath)
-	err = generateOutputFile(codeowners, outputPath, opts, cmd)
+	opts.logger.V(logging.LogDebug).Style(0, colors.FgBlue).Infof("Processing codeowners file at: %s\n", opts.outputPath)
+
+	err = generateOutputFile(codeowners, filepath.Join(opts.outputPath, fileType), opts, cmd)
 	if err != nil {
 		_ = opts.telemetry.CaptureFailedCodeownersGenerate()
 		return fmt.Errorf("error generating github style codeowners file: %w", err)
 	}
-	opts.logger.V(logging.LogInfo).Style(0, colors.FgGreen).Infof("Finished generating file: %s\n", outputPath)
+
+	opts.logger.V(logging.LogInfo).Style(0, colors.FgGreen).Infof("Finished generating file: %s\n", filepath.Join(opts.outputPath, fileType))
 	_ = opts.telemetry.CaptureCodeownersGenerate()
 
 	opts.logger.V(logging.LogInfo).Style(0, colors.FgCyan).Infof("\nCreate an OpenSauced Contributor Insight to get metrics and insights on these codeowners:\n")
--- a/cmd/generate/codeowners/output.go
+++ b/cmd/generate/codeowners/output.go
@@ -15,6 +15,15 @@ import (
 )
 
 func generateOutputFile(fileStats FileStats, outputPath string, opts *Options, cmd *cobra.Command) error {
+
+	// Create specified output directories if necessary
+	err := os.MkdirAll(filepath.Dir(outputPath), os.ModePerm)
+	if err != nil {
+		if !os.IsExist(err) {
+			return fmt.Errorf("error creating directory at %s filepath: %w", outputPath, err)
+		}
+	}
+
 	// Open the file for writing
 	file, err := os.Create(outputPath)
 	if err != nil {
--- a/docs/pizza_generate_codeowners.md
+++ b/docs/pizza_generate_codeowners.md
@@ -38,6 +38,9 @@ pizza generate codeowners . --owners-style-file
 
 # Specify a custom location for the .sauced.yaml file
 pizza generate codeowners . --config /path/to/.sauced.yaml
+
+# Specify a custom output location for the CODEOWNERS file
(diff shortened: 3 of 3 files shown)


cmd/generate/codeowners/codeowners.go
```diff
@@ -176,21 +181,24 @@ func run(opts *Options, cmd *cobra.Command) error {
 		return fmt.Errorf("error traversing git log: %w", err)
 	}
 
-	// Bootstrap codeowners
-	var outputPath string
+	// Add optional output path to base path
+	fullOutputPath := filepath.Join(opts.path, opts.outputPath)
```
jpmcb: This creates some unexpected behavior and should be the fully resolved path of whatever the user's provided output path is. For example, if I give the following command:

```
❯ pizza generate codeowners . -l debug -o /tmp

Finished generating file: /Users/jpmcb/workspace/opensauced/pizza-cli/tmp/CODEOWNERS
```

I can see that it generated a file at `./tmp/CODEOWNERS` and not the root `/tmp` like I provided.

cmd/generate/codeowners/codeowners.go
```diff
@@ -176,21 +181,24 @@ func run(opts *Options, cmd *cobra.Command) error {
...
+
+	// Define the file type based on a flag
+	var finalOutputPath string
```
jpmcb: Why the var name change from `outputPath` to `finalOutputPath`?

jpmcb (changes requested):

cmd/generate/codeowners/codeowners.go
udaypatel1: I see –

From my understanding, I was assuming that the output directory would likely be wanted under the root of the current repository.

I'll work on generalizing this to align with what you pointed out. I'll aim to follow similar logic to that of the `-o` flag for `pizza generate config`

cmd/generate/codeowners/codeowners.go
udaypatel1: Initially, this was to differentiate between the `fullOutputPath` variable.

As I'll go through a rewrite of parsing the `outputPath` properly and not assuming it is under the root of the current repository, this will be removed.

cmd/generate/codeowners/codeowners.go
jpmcb: > From my understanding, I was assuming that the output directory would likely be wanted under the root of the current repository.

That's one way to approach it. But, typically, when I've seen a `-o` output flag used, it's a fully qualified path. So, if you wanted to drop it in a `.bitbucket` directory, you'd just add the `./` to make it a fully resolvable path:

```sh
pizza generate codeowners . -o ./.bitbucket
```

but in my case, if I provide a full path (like `/tmp` or `~/.config/pizza-cli`), it should resolve to that path.

jpmcb:
Great work on this! Thanks for taking the initiative to pick this up!! 👏🏼👏🏼 👏🏼

cmd/generate/codeowners/codeowners.go
udaypatel1: @jpmcb – pushed a change; this functionality should now align with your expectations above.

Very similar logic to what @zeucapua used for the `-o` flag on `pizza generate config`

Do let me know if you'd like me to make this DRY and abstract it further, unless that's out of scope for this PR.

cmd/generate/codeowners/codeowners.go
```diff
@@ -139,6 +143,7 @@ pizza generate codeowners . --config /path/to/.sauced.yaml
 
 	cmd.PersistentFlags().IntP("range", "r", 90, "The number of days to analyze commit history (default 90)")
 	cmd.PersistentFlags().Bool("owners-style-file", false, "Generate an agnostic OWNERS style file instead of CODEOWNERS.")
+	cmd.PersistentFlags().StringP("output-path", "o", "./", "Directory to create the output file.")
```
jpmcb: If the default is `./` then when someone runs this from _outside_ the target directory (which is specified by the `path` argument), this will land it in the current directory which is unexpected:

```
❯ pizza generate codeowners /some/target/directory
❯ ls -la ./CODEOWNERS
-rw-r--r--@ 1 jpmcb  staff  4954 Oct 15 11:15 ./CODEOWNERS
```

I'd imagine the default here should be the target directory: not just the current one

cmd/generate/codeowners/codeowners.go
```diff
@@ -176,21 +181,33 @@ func run(opts *Options, cmd *cobra.Command) error {
...
+	} else {
+		err := generateOutputFile(codeowners, filepath.Join(opts.outputPath, fileType), opts, cmd)
+		if err != nil {
+			_ = opts.telemetry.CaptureFailedCodeownersGenerate()
+			return fmt.Errorf("error generating github style codeowners file: %w", err)
+		}
 	}
-	opts.logger.V(logging.LogInfo).Style(0, colors.FgGreen).Infof("Finished generating file: %s\n", outputPath)
+
+	opts.logger.V(logging.LogInfo).Style(0, colors.FgGreen).Infof("Finished generating file: %s\n", opts.path)
```
jpmcb: This looks to be a mistake since we don't want to log the path but rather the output path:

```suggestion
	opts.logger.V(logging.LogInfo).Style(0, colors.FgGreen).Infof("Finished generating file: %s\n", opts.outputPath)
```

cmd/generate/codeowners/codeowners.go
```diff
@@ -176,21 +181,33 @@ func run(opts *Options, cmd *cobra.Command) error {
...
-	opts.logger.V(logging.LogDebug).Style(0, colors.FgBlue).Infof("Processing codeowners file at: %s\n", outputPath)
-	err = generateOutputFile(codeowners, outputPath, opts, cmd)
-	if err != nil {
-		_ = opts.telemetry.CaptureFailedCodeownersGenerate()
-		return fmt.Errorf("error generating github style codeowners file: %w", err)
+	opts.logger.V(logging.LogDebug).Style(0, colors.FgBlue).Infof("Processing codeowners file at: %s\n", opts.path)
+
+	// Generate file at a specified location based on a flag or at the default path
+	if opts.outputPath == "~/" {
+		homeDir, _ := os.UserHomeDir()
```
jpmcb: I'm guessing we weren't handling the home directory resolution correctly?

jpmcb (changes requested):

cmd/generate/codeowners/codeowners.go
udaypatel1: That makes sense–

Made the following changes:

1) Default the `outputPath` argument to be the `path` argument (target directory). This will ensure that if the `outputFlag` is not supplied, the base location will be at the root of the repo path which is a mandatory argument

2) If the `outputPath` is provided, it will generate the file in that specified location, distinct from the repo path argument. If that `outputPath` directory does not exist, the directory (and any subdirectories) provided will be created. Also, the `outputPath` is a fully qualified location.

Do let me know if I'm missing any other nuances!

cmd/generate/codeowners/codeowners.go
udaypatel1: This logic was used in `pizza generate config -o /path/to/directory`, I had left it unchanged.

This flow has since been removed on my updated push.

udaypatel1:
Log from within the target repo + flag:

image

Log from outside of the target repo + flag:

image

Log from within the target repo + no flag:

image

Log from outside of the target repo + no flag:

image

---------------------

Feel free to suggest any log changes that you may want, if any.

cmd/generate/codeowners/codeowners.go
```diff
@@ -176,21 +190,23 @@ func run(opts *Options, cmd *cobra.Command) error {
...
-		outputPath = filepath.Join(opts.path, "OWNERS")
+		fileType = "OWNERS"
 	} else {
-		outputPath = filepath.Join(opts.path, "CODEOWNERS")
+		fileType = "CODEOWNERS"
 	}
 
-	opts.logger.V(logging.LogDebug).Style(0, colors.FgBlue).Infof("Processing codeowners file at: %s\n", outputPath)
-	err = generateOutputFile(codeowners, outputPath, opts, cmd)
+	opts.logger.V(logging.LogDebug).Style(0, colors.FgBlue).Infof("Processing codeowners file at: %s\n", opts.path)
```
jpmcb:
```suggestion
	opts.logger.V(logging.LogDebug).Style(0, colors.FgBlue).Infof("Processing codeowners file at: %s\n", opts.outputPath)
```

jpmcb (approved):

nickytonline (approved):
Thanks for this @udaypatel1!  🚢

